PR #4, opened 2019-04-16
Res 910

subutai:
@lscheinkman Note the change to model_utils and mnist_ray


projects/ray_tune_example/mnist/mnist_ray.py
```diff
@@ -246,6 +246,10 @@ def parse_options():
   projectDir = os.path.dirname(options.config.name)
   projectDir = os.path.abspath(projectDir)
 
+  # Pre-download dataset
+  datadir = os.path.join(projectDir, "data")
+  train_dataset = datasets.CIFAR10(datadir, download=True, train=True)
```
lscheinkman: I think you mean MNIST here

PR #10, opened 2019-05-08
RES-913: MobileNet CIFAR10 experiments

lscheinkman:
@subutai Here are the MobileNet experiments with CIFAR10.
I will add noise test in a separate PR


projects/whydense/cifar/mobilenet_cifar.py
```diff
@@ -0,0 +1,190 @@
+#  Numenta Platform for Intelligent Computing (NuPIC)
```
subutai: Shouldn't this be in frameworks/pytorch?

lscheinkman: I dont know. It seems to specialized to be used by some other project.

PR #29, opened 2019-06-10
Res 942

onejgordon:
Experiment with weight initialization (LSUV & Grassmannian) on CIFAR10.


projects/weight_init/weight_init_experiment.py
```diff
@@ -0,0 +1,671 @@
...
+
+    def initialize(self):
+        for m in self.model.modules():
+            if isinstance(m, nn.Conv2d):
+                n_filters = m.out_channels
+                n_input = m.in_channels
+                k = m.kernel_size[0]
+                w = self.grassmannian_extract(n_filters, k * k, n_input)
+                m.weight.data = torch.from_numpy(w).float().to(self.device)
+                # Break after first Conv layer
```
subutai: Why only the first layer?

projects/weight_init/weight_init_experiment.py
```diff
@@ -0,0 +1,671 @@
...
+        self.linear_weight_sparsity = config["weight_sparsity"]
+        self.linear_n = config["linear_n"]
+        self.linear_percent_on = config["linear_percent_on"]
+        if isinstance(self.linear_n, int):
+            self.linear_n = [self.linear_n]
+            self.linear_percent_on = [self.linear_percent_on]
+            self.linear_weight_sparsity = [self.linear_weight_sparsity]
+        self.output_size = config.get("output_size", 10)
+
+        self.weight_init = config.get("weight_init")
```
subutai: Maybe this should default to "default"?

projects/weight_init/experiments.cfg
```diff
@@ -0,0 +1,107 @@
...
+cnn_out_channels = [64, 128, 256, 512, 512]
+cnn_kernel_size = [3, 3, 3, 3, 3]
+linear_n = []
+linear_percent_on = []
+weight_sparsity = []
+
+stop = {"stop": 1}
+checkpoint_at_end = True
+
+#[VGG19Dense]
```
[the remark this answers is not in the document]
onejgordon: Ah yes, now seeing that's what we use the --experiment arg for.

projects/weight_init/weight_init_experiment.py
onejgordon: Prabhu's paper introducing Grassmannian used pre-packed weights only for the first layer: 
"With the intuition that the first few convolutional kernelfilters should capture as much diverse features as possibleto prevent ‘dead kernels’ (Zeiler & Fergus, 2014)..."

May be interesting to try with all layers as well.

projects/weight_init/weight_init_experiment.py
onejgordon: Good point. I wasn't clear on how we balance defaults in the experimental cfg default section vs the code.

PR #31, opened 2019-06-18
Cifar100 merge

lucasosouza:


projects/whydense/cifar-100/cifar_experiment.py
```diff
@@ -20,29 +20,30 @@
...
-    SparseWeights,
-    SparseWeights2d,
     Flatten,
-    KWinners2d,
     KWinners,
+    KWinners2d,
+    SparseWeights,
+    SparseWeights2d,
     rezeroWeights,
     updateBoostStrength,
```
lscheinkman: @lucasosouza `nupic.torch` API has changed to be pep8 compliant and these functions were renamed.
See:
 - https://github.com/numenta/nupic.torch/blob/master/nupic/torch/modules/k_winners.py#L31 
 - https://github.com/numenta/nupic.torch/blob/master/nupic/torch/modules/sparse_weights.py#L29

lscheinkman: You should get the latest `nupic.torch` and fix the API changes before merging

lucasosouza: Modified and tested with new nupic.torch, please review

PR #32, opened 2019-06-18
model utils bug fix

subutai:
Fixes off by one issue with batches in epoch. e.g., with batches_in_epoch==1 it will now just run one batch.


projects/whydense/cifar/cifar_run.py
```diff
@@ -34,7 +34,7 @@
 def train_models(configs, project_dir):
     """Run all the training experiments specified in configs."""
     # download dataset
-    data_dir = os.path.join(project_dir, "data")
+    data_dir = os.path.join(os.environ["HOME"], "nta", "data")
```
lscheinkman: "nta" is specific to our laptops, it may not work everywhere.

nupic/research/frameworks/pytorch/model_utils.py
```diff
@@ -127,15 +127,14 @@ def evaluate_model(
 
     with torch.no_grad():
         for batch_idx, (data, target) in enumerate(loader):
+            if batch_idx >= batches_in_epoch:
+                break
```
lscheinkman: Should `train_model` do the same ?

subutai: yes, I think so. Pushing...

projects/whydense/cifar/cifar_run.py
subutai: It's also on AWS. This way it's convenient since the exact paths are different between my laptop and AWS. Also, it's only the default if you don't specify in config.  I pushed a change to make it clearer.

PR #35, opened 2019-06-24
RES-970: Not So Dense

lscheinkman:
@subutai Please review.


lscheinkman:
@subutai Added sparse weights option

PR #42, opened 2019-07-16
added boosting

lucasosouza:
Ongoing experiment version, not fit for production. Added:
- DSNN model. Started experimenting new modifications, such as decay, flip, gradient based pruning. More features to be added
- Added KWinneers from nupic.torch to the model. Sample Network (VGG) being used is similar to whydense
- Added sample implementation of KWinners, using kthvalue, with issues to be fixed. Not being used as of now

### DIFF
--- a/projects/dynamic_sparse/functions.py
+++ b/projects/dynamic_sparse/functions.py
@@ -1,0 +1,106 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2019, Numenta, Inc.  Unless you have an agreement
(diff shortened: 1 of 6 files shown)


projects/dynamic_sparse/functions.py
```diff
@@ -0,0 +1,106 @@
...
+        k = int(self.k_perc * num_units)
+
+        # initialize duty cycle
+        if self.duty_cycle is None:
+            # for MLP alone
+            self.duty_cycle = torch.zeros(units_shape)
+
+        # keep track of number of past iteratctions
+        if self.current_time < self.T:
+            self.current_time += 1
```
lscheinkman: You should add the batch size here not just one

PR #44, opened 2019-07-18
RES-979: Port k-winners to tensorflow

lscheinkman:
@onejgordon @subutai Please review. 
I've also ported the new tests from nupic.torch to tensorflow


nupic/research/frameworks/tensorflow/constraints/__init__.py
```diff
@@ -0,0 +1,21 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2018, Numenta, Inc.  Unless you have an agreement
```
subutai: 2019?

nupic/research/frameworks/tensorflow/constraints/sparse_weights.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Constrains the weights to a fixed sparsity rate where a fixed number of
+    weights are always zeros.
+
+    :param sparsity:
+      Percentage of weights that are allowed to be non-zero. Default 0.5
+    :type sparsity: float
+    """
+
+    def __init__(self, sparsity=0.5, name=None):
```
subutai: Should we call this percent_on instead of sparsity?

lscheinkman: Good point, should we also rename the pytorch version?
https://github.com/numenta/nupic.torch/blob/master/nupic/torch/modules/sparse_weights.py#L105

onejgordon: This is fundamentally different than the percent_on arg in KWinners right? If so, I see some value in using unambiguous terms.

subutai: Yes, whatever term we use, we should make the usage clear. We could use percent_non_zero if that is clearer.

I've found that the word sparsity is really tough for people once we get to specifics. Does sparsity=0.2 mean 20% are non-zero or that 80% are non-zero?

You could argue the current usage is incorrect:
`        :param weight_sparsity:
          Pct of weights that are allowed to be non-zero in the layer.
`

lscheinkman: I will rename `sparsity ` to `percent_non_zero` on `nupic.research` tensorflow version only for now. 
The `nupic.torch` change is a bigger change involving multiple repositories. I will do that next week.

PR #47, opened 2019-08-15
Speech experiment mods

subutai:


subutai:
@lscheinkman Please see dataset_utils.py  and gsc/sparse_speech_experiment.py. You can ignore model_compare

PR #48, opened 2019-08-19
Remove batchnorm script

subutai:


subutai:
To use it:
        `model = gsc_sparse_cnn(pretrained=True)`
        `model2 = remove_batchnorm(model)`

subutai:
@lscheinkman Could definitely use some help making the core function `remove_batchnorm` more readable.

PR #54, opened 2019-09-05
Move stuff to frameworks

subutai:


subutai:
@lucasosouza Pycharm has some easy to use refactor tools, so I used it to move the files around as discussed.

subutai:
Without these changes I can't run it on my machine.

subutai:
> maybe we should set the folder where data is saved and results are saved to be environment variables instead? It will solve the problem of everyone using a different folder

Sure, let's do that in a separate PR. There are a few similar small things that might make life a bit easier. Currently it's still pretty flexible: the run files can always include `data_dir=os.environ['DATADIR']` if someone wants it.

PR #60, opened 2019-09-12
Dynamic sparse tests

lucasosouza:


lscheinkman:
@lucasosouza The tests didnt run. We need to update the circleci script before we can merge

PR #65, opened 2019-09-13
K winners

subutai:


subutai:
> Would you mind if I pass the index to the block, and then to kwinners function, and kwinners will look for the required args in config? The goal was to keep the network initialization agnostic to the activation function. I can make the changes after merging if you agree.

You mean an index into _linear_block() right? I think that will be cleaner - I can make the change.

subutai:
> We can't move post epoch updates to before evaluation, because pruning is done there. Pruning has to be done after evaluation. Maybe a different method, if it is required?

Sounds good - I'll make a suggested change.

subutai:
Made the changes, but it's dependent on https://github.com/numenta/nupic.torch/pull/43 so will wait until that is approved.

PR #68, opened 2019-09-13
Update installation instructions

lscheinkman:
@subutai please review


lscheinkman:
only if we have problems, 
The last line in environement.yml will install nupic.research in develop mode and any other dependency not listed in the yml will be installed from pipy like today

PR #71, opened 2019-09-19
Refactored sparse CNN creation a bit

subutai:


tests/unit/frameworks/pytorch/le_sparse_net_test.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_no_cnn(self):
+        """Create a net where there are no CNN blocks."""
+        model = LeSparseNet(
+            cnn_out_channels=(),
+            linear_n=(100, 200),
+            linear_activity_percent_on=(0.1, 1.0),
+            linear_weight_percent_on=(1.0, 0.4),
+        )
+        self.assertGreater(len(model._modules), 2)
+        for key in model._modules.keys():
```
lscheinkman: You should use `named_modules()` instead. Something like this:
```
for key, val in model.named_modules():
```

tests/unit/frameworks/pytorch/le_sparse_net_test.py
```diff
@@ -0,0 +1,97 @@
...
+        self.assertEqual(y.size()[1], 10)
+
+    def test_no_linear(self):
+        """Create a net where there are no linear blocks."""
+        model = LeSparseNet(
+            cnn_out_channels=(8, ),
+            linear_n=(),
+        )
+        self.assertGreater(len(model._modules), 2)
+        for key in model._modules.keys():
```
lscheinkman: Replace with `named_modules()`. see previous comment

PR #77, opened 2019-10-03
Extra logging + new options for dynamic-sparse modules

mvacaporale:
The changes included mostly serve either one of two purposes
* Optional logging of coactivations vs weight-magnitudes 
* Optional rules for calculating coactivations: using non-binary activations and accumulating a running exponential mean of coactivations instead of sum.

For the former, there's a new ability for `loggers.py` to  handle configurable scatter plots.


nupic/research/frameworks/dynamic_sparse/models/main.py
```diff
@@ -519,13 +525,53 @@ def _run_one_pass(self, loader, train, noise=False):
                 self.log["val_loss"] = loss
                 self.log["val_acc"] = acc
 
+        # save scatter plot of magn vs coacts.
+        if train and self.log_magnitude_vs_coactivations:
+            for i, m in enumerate(self.dynamic_sparse_modules):
+
+                coacts = m.coactivations.clone().detach().to("cpu").numpy()
```
lucasosouza: Maybe more clean to use .cpu() instead of .to("cpu")?

nupic/research/frameworks/dynamic_sparse/models/main.py
```diff
@@ -519,13 +525,53 @@ def _run_one_pass(self, loader, train, noise=False):
...
+                weight = m.weight.clone().detach().to("cpu").numpy()
+                grads = m.weight.grad.clone().detach().to("cpu").numpy()
+
+                mask = ((m.weight.grad != 0)).to("cpu").numpy()
+
+                coacts = coacts[mask]
```
lucasosouza: wouldn't it be more efficient to take to cpu() after masking?

nupic/research/frameworks/dynamic_sparse/networks/layers.py
```diff
@@ -164,7 +180,15 @@ def reset_coactivations(self):
...
-        self.coactivations[:] += new_coacts
+        if self.update_func:
+            updated_coacts = self.update_func(self.coactivations, new_coacts)
+        elif self.moving_average_alpha:
+            alpha = self.moving_average_alpha
+            updated_coacts = (1 - alpha) * self.coactivations + alpha * new_coacts
+        else:
+            updated_coacts = self.coactivations + new_coacts
+
+        self.coactivations[:] = updated_coacts
```
lucasosouza: In the case of moving average, should we still reset it after each epoch? Maybe we can take that off in the case of moving average and move from the concept of "epoch" to just iterations.

PR #94, opened 2019-10-12
remove dependencies from specific types of layers in sparse model and…

lucasosouza:
Removes dependencies on specific type of layers from modules, restoring the flexibility of models to work with different types of networks and implementations. Goal is to restore compatibility to other implementations rather than DSNN, as well as previous and future versions of DSNN that do not make use of `DynamicSparseBase` layers, without disrupting current experiments. No changes were done in hebbian.py or in layers.py, only in the models. Current code can still run without any changes.

Working version, under thorough test (running runs/test_restoration.py). Changes might be added in the future to clean/simplify code without affecting the API. 

Note: might break some of the newly added custom logging. Logging will be refactored in the next PR. 

Details:
- New wrapper class `SparseModule`, belonging to model. Should contain all attributes required for pruning, growing and inspecting, and common operations pertaining the module such as initialize mask or apply mask.
- Removed all dependencies to network specific implementations, such as custom layers
- Only one list containing sparse modules. If more information is required, such as making a specific sparse layer dynamic, can add a new attribute to the SparseModule class. Removed all references to `dynamic_sparse_layers`.
- Refactored prune code to operate on `SparseModule`
- Restored hebbian_v0 approach into a separate file, as its original version with few modifications to conform to the new API.
- Implemented attribute `sparse_type`: can be `precise` (default), which guarantees the exact sparsity per layer; `precise_per_output`, same but guarantees sparsity per output neuron or channel in case of conv; or `approximation`, which is computationally faster.
- No changes applied to hebbian.py, layers.py or related. Only the models were affected.

Test results:
image


nupic/research/frameworks/dynamic_sparse/models/dsnn.py
```diff
@@ -39,39 +31,48 @@ class DSNNHeb(SparseModel):
...
-        """
-        self.network.apply(init_coactivation_tracking)
+    def _init_hebbian(self):
+        for module in self.sparse_modules:
+            if self._is_dynamic(module):
+                module.init_coactivation_tracking()
+        if hasattr(self.network, "forward_with_coactivations"):
+            self.network.forward = self.network.forward_with_coactivations
+
+    def _is_dynamic(self):
```
mvacaporale: You have `_is_dynamic` take either 1 input (`self`) or two inputs (`self` and `module`), but you should keep its arguments consistent.

nupic/research/frameworks/dynamic_sparse/models/dsnn.py
```diff
@@ -39,39 +31,48 @@ class DSNNHeb(SparseModel):
...
-        """
-        Override method in children classes
-        Should only track coactivations if required by the algorithm
-        """
-        self.network.apply(init_coactivation_tracking)
+    def _init_hebbian(self):
+        for module in self.sparse_modules:
+            if self._is_dynamic(module):
+                module.init_coactivation_tracking()
+        if hasattr(self.network, "forward_with_coactivations"):
```
mvacaporale: In the future, I believe we shouldn't have to worry about dealing with networks with a `forward_with_coactivations` (Thanks for leaving this a flexible option). I think it will be a lot easier to introspect a network, find which modules deserve coactivations, and then track them without the network knowing.

I'm not a 100% sure about the feasibility of this, but it would enable us to be truly agnostic to the network. In contrast, I'm wondering whether a `forward_with_coactivations` will enable the same flexibility. If anything, it may just be cumbersome overhead to achieve what forward hooks can achieve.

nupic/research/frameworks/dynamic_sparse/networks/hebbian_v0.py
```diff
@@ -0,0 +1,252 @@
...
+        """A faster and approximate way to track correlations"""
+
+        if "features" in self._modules:
+            x = self.features(x)
+
+        prev_act = (x > 0).detach().float()
+        for idx_layer, layer in enumerate(self.classifier):
+            # do the forward calculation normally
+            x = layer(x)
+            if self._has_params(layer) and hasattr(layer, "coactivations"):
```
mvacaporale: As far as I can tell, it doesn't seem as if you're tracking the coactivations: `hasattr(layer, "coactivations")` will always come up false as you never define them for a given module.

This is only a superficial observation - I haven't dived deep in the code. I may be wrong, but I just want to make sure you're certain that you're tracking coactivation as you intend.

lucasosouza: Coactivations is being initialized at `init_coactivation_tracking` in SparseModule

nupic/research/frameworks/dynamic_sparse/models/dsnn.py
lucasosouza: Thanks. Fixed, will send another commit.

nupic/research/frameworks/dynamic_sparse/models/dsnn.py
lucasosouza: Agreed. This is temporary for backwards compatibility. I do intend to work with the hooks approach you proposed.

PR #97, opened 2019-10-17
refactors log out of models into a separate class

lucasosouza:
Refactors all logging functions into a separate class. Goal is to clean  code inside model and separate responsibilities - code required to run the model and log the variables should be separate. 

Model calls logger at a few keypoints, such as pre and post epoch, to trigger a logging event. Loggers have to access to model and may inspect (but not modify) its internals. Custom logging functions may also pass additional variables.

Added:
- new file loggers.py, in models folder for now, including: BaseLogger, SparseLogger(BaseLogger), and DSNNLogger(SparseLogger)
- changed loggers.py in util to ray_custom_loggers.py for differentiation


nupic/research/frameworks/dynamic_sparse/models/main.py
```diff
@@ -310,10 +278,12 @@ def setup(self):
                     module.apply_mask()
                     module.save_num_params()
 
-        # TODO: remove logging to separate class
+        self.logger = SparseLogger(self)
+        # TODO: is this still required?
```
lucasosouza: @mvacaporale Not sure what this was supposed to do, I didn't get the logic here. Can you please check if it is still required?

nupic/research/frameworks/dynamic_sparse/models/loggers.py
```diff
@@ -0,0 +1,220 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
```
subutai: Should this be in the models directory, or in common or even loggers directory?

lucasosouza: I am also not sure. I've left there temporarily while I figure out a better place, I will take suggestions. Probably in common

subutai: 👍

nupic/research/frameworks/dynamic_sparse/models/loggers.py
```diff
@@ -0,0 +1,220 @@
...
+import numpy as np
+import torch
+from pandas import DataFrame
+
+
+class BaseLogger:
+    def __init__(self, model, config=None):
+        defaults = dict(debug_weights=False, verbose=0)
+        defaults.update(config or {})
+        self.__dict__.update(defaults)
```
mvacaporale: Small note: I've thought about the method of setting class attributes via the config a bit, and there's potential edge case which would should avoid - overwriting pre-existing attributes. As well, I think we should avoid adding more attributes than necessary. If config has more params than just `debug_weights` and `verbose`, we'd not only update those attrs, but we also add other params and potentially change the value of some pre-existing keys in `self.__dict__`. 

Something like the following would work
```
        config = config or {}
        defaults = dict(
            debug_weights=False,
            verbose=0,
        )
       # Only update desired defaults. 
        new_defaults = {k: (config.get(k, None) or v) for k, v in defaults.items()}
        self.__dict__.update(new_defaults)
```

nupic/research/frameworks/dynamic_sparse/models/main.py
mvacaporale: Which line are you talking about? The one with `log_magnitude_vs_coactivations`?

lucasosouza: The line 282 to 284 there

nupic/research/frameworks/dynamic_sparse/models/loggers.py
lucasosouza: It is implemented that way specifically to override attributes (e.g. a new class can set a new default to a variable, different than the parent class). But I did not think through this a lot, this was the quick and dirty solution, we can come up with something better if that is not working.

PR #98, opened 2019-10-17
Le sparse updates

subutai:


projects/whydense/gsc/sparse_speech_experiment.py
```diff
@@ -140,7 +142,8 @@ def __init__(self, config):
 
     def save(self, checkpoint_path):
         checkpoint_path = os.path.join(checkpoint_path, "model.pt")
-        torch.save(self.model.state_dict(), checkpoint_path)
+        if self.last_training_epoch >= self.min_epoch_for_checkpoint:
+            torch.save(self.model.state_dict(), checkpoint_path)
```
lscheinkman: **`ray tune`** has a few parameters to configure how often it checkpoints.
https://ray.readthedocs.io/en/latest/tune-usage.html#trainable-trial-checkpointing

subutai: Thanks - will look into that.

PR #104, opened 2019-10-18
Res1140 testnoise

lucasosouza:
Runs noise evaluation on ResNet and WideResNet

- Adds save and restore capabilities to base model
- Adds script to test noise after the model has ran
- Moves test noise in datasets to a separate function, so it can run independently of the model
- Run file to replicate HSD experiments on CIFAR-10 using Resnet and WideResnet (testing with vs without kwinners, and weight dense vs weight sparse)
- Minor fixes and additional comments to browser.py to address bugs

For results, see:
https://ntaresearch.atlassian.net/secure/RapidBoard.jspa?rapidView=60&modal=detail&selectedIssue=RES-1140&assignee=5cc8ca4d9e366e0e6582a383


projects/dynamic_sparse/runs/resnet_cifar.py
```diff
@@ -0,0 +1,73 @@
...
+    dropout_rate=0,  # zero dropout in wideresnet
+    # ---- optimizer related
+    optim_alg="SGD",
+    learning_rate=0.1,
+    lr_scheduler="MultiStepLR",
+    # lr_milestones=[60, 90, 110],
+    lr_milestones=[60, 120, 160],
+    lr_gamma=0.2,
+    weight_decay=0.0005,
+    momentum=0.9,
```
subutai: BTW, I have quite often found that momentum hurts sparse networks. It makes sense - I can explain it in person. We should use 0 momentum for sparse nets.

lucasosouza: Thanks. I did see that happen in GSC, but I thought it was a particularity of GSC networks and not of sparse networks. I can see how momentum wouldn't make a lot of sense in dynamic sparse (connection) networks, but I don't see that same case in the static sparse (connection) networks.

PR #106, opened 2019-10-20
Res 1145 pruning schedule

lucasosouza:
Implements pruning schedule. At any point during training, the weight sparsity can be set and the model will be pruned or grown accordingly.

Changes:
- separates prune and grow into different methods
- moves SparseModule into a separate modules file (modules.py)
- create a new model specifically for pruning (pruning.py)


nupic/research/frameworks/dynamic_sparse/models/main.py
```diff
@@ -185,25 +191,42 @@ def has_params(self, module):
...
+        If path is a directory, will append the parameter model_filename
+        """
+        # print("loading from", checkpoint_path)
+        # experiment_root = os.path.join(checkpoint_dir, experiment_name)
+        checkpoint_path = os.path.join(checkpoint_dir, experiment_name + ".pth")
+        self.network.load_state_dict(
+            torch.load(checkpoint_path, map_location=self.device)
+        )
+
+    def evaluate_noise(self, dataset):
```
mvacaporale: Not every model may require a noise dataset correct? It could be good to include logic that covers such cases.

lucasosouza: We discussed, but just to keep the record, the dataset is passed to the function so that should not be an issue.

PR #107, opened 2019-10-21
RES-985: Some random but minor enhancements. 

mvacaporale:
A few fixes, some minor additions, and some tidying up of the directories:

- Make sure the config is getting passed to the loggers.
- Adjust implemetation of `_NullConv` class so it's not registered as a `sparsifiable` module.
- Organize my own `run` and `notebook`files into seperate sub-directory.
- Add implementation for a small dense GSC network - this is configurable for a desired `on_percentage`.


nupic/research/frameworks/dynamic_sparse/models/main.py
```diff
@@ -65,6 +66,7 @@ def __init__(self, network, config=None):
         # init remaining
         self.device = torch.device(self.device)
         self.network = network.to(self.device)
+        self.config = deepcopy(config)
```
lucasosouza: I see what you mean now by overriding. Cool, this looks like a good solution for now

mvacaporale: My `overriding` comment was about something else actually, but I think I was mistaken so you can forget I mentioned anything. 

The addition above was to help fix a bug with the loggers: The `self.config = deepcopy(config)` allows the model to save a copy of the config during the `__init__`, so that it may pass the config to its logger during that `setup` - as opposed to before when the loggers were being initialized without any config and therefore using only their defaults (e.g `debug_weights=False`)

PR #151, opened 2019-12-03
added extra tests to ensure compatibility with torchvision resnet50

lucasosouza:


tests/unit/frameworks/pytorch/resnets_test.py
```diff
@@ -235,6 +238,58 @@ def test_different_sizes(self):
 
         self.assertIsInstance(net, ResNet, "Loads ResNet18")
 
+    def test_version_15(self):
+        """Evaluates if dense resnet-50 is similar to torchvision resnet-50"""
```
lscheinkman: I am not sure this test is necessary. 
The real test should be against `resnet50 v1.5` not the specific pytorch implementation. 
Pytorch may decide to change their implementation (i.e. "layer names") and break this test.

lucasosouza: You are right, also not sure if the test is necessary later down the road, but I thought it was important to ensure we don't deviate from the 1.5 implementation by accident. I would highlight it is not specific to torchvision implementation but would work with any implementation of ResNet 1.5 in pytorch. The layer names are not given names, but rather the class name, such as Conv2d or Batch Norm - that test merely ensures that all the layers exist and are in the right order. The test would only fail if pytorch changes the name of the class, say from Conv2d to Convolutional2d for example, but that would mean we would have to change it in our code as well since we are importing from pytorch.

lscheinkman: This version looks like it is making sure our `resnet50` implementation matches `torchvision.models.resnet50` implementation not necessary "resnet50 v1.5" definition.
This test will fail when torchvision decides to change their default `torchvision.models.resnet50()` implementation to v1 or some other version down the road.
If the goal of the test is to make sure our implementation matches resnet50 v1.5 then we should validate against the definition of "resnet50 v1.5" and not depend on torchvision default implementation otherwise we are just testing compatibility with the current version in torchvision.
However if the goal of this test is make sure our implementation matches torchvision's implementation then rename to test to reflect that goal, something like `def test_match_torchvision_version(self)`

PR #170, opened 2020-01-10
Added custom auto params to experiments file

subutai:


projects/imagenet/experiments_custom_super.py
```diff
@@ -0,0 +1,129 @@
...
+
+    if percent_on >= 1.0:
+        return None
+    else:
+        return dict(
+            percent_on=percent_on,
+            boost_strength=1.0,
+            boost_strength_factor=0.9,
+            local=True,
+        )
```
mvacaporale: Be sure to add the `k_inference_factor=1.0` to the return dict here.

PR #201, opened 2020-02-18
Validation, timing refinements

subutai:


nupic/research/frameworks/pytorch/model_utils.py
```diff
@@ -91,16 +91,17 @@ def train_model(
...
         num_images = len(target)
-        t1 = time.time()
         if batch_idx >= batches_in_epoch:
             break
         if pre_batch_callback is not None:
             pre_batch_callback(model=model, batch_idx=batch_idx)
 
         data = data.to(device, non_blocking=async_gpu)
         target = target.to(device, non_blocking=async_gpu)
+        t1 = time.time()
```
lscheinkman: You are combining "data loading" time with "pre_batch_callback" time and "cpu/gpu" memory transfer time with "t1"

subutai: Yes that's true, but currently that callback doesn't do anything. I wanted to include the .to() call but that may not do much - should I just move t1 to be earlier instead?

subutai:
@lscheinkman I changed the order a bit in model_utils to address your point.

nupic/research/frameworks/pytorch/model_utils.py
lscheinkman: > should I just move t1 to be earlier instead?

It depends what is being profiled. 
I was just pointing out that if you include the ".to()" timing to "t1" then "t1" represents "loading data from disk/memory" and  "moving data from cpu to gpu". If you assume "moving data to gpu" does not impact your stats then it is ok. 
When i used pytorch profiler, i found out that most of the time was spent on moving data from cpu to gpu via ".item()", that's why "validate()" is still slow.

PR #212, opened 2020-02-29
Sigopt utilities

subutai:


subutai:
Very experimental code. Main thing is to make sure imagenet_tune is not messed up somehow.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -295,6 +310,15 @@ def _train(self):
         # Wait for remote functions to complete
         results = ray.get(status)
 
+        # Update the sigopt configuration once we're at the end
+        if self.iteration >= self.epochs - 1 and self.sigopt is not None:
```
lscheinkman: This assumes the stop criteria to always be `epochs` and may not work with early stop or other stop criteria.
Maybe you could save the latest accuracy value here and put the update logic in the `_stop` method instead?

PR #216, opened 2020-03-06
Sigopt updates

mvacaporale:
This PR includes a few changes to the sigopt related utilities for added flexibility when running experiments.

Changes include:
* Specification of `<SigoptExperiment>` class from config. Thus `SigOptImagenetExperiment` may be subclassed by the user as they see fit.
* The functionality to enable multiple metrics (See `SigOptImagenetTrainable._process_result`)


nupic/research/frameworks/sigopt/sigopt_experiment.py
```diff
@@ -251,3 +251,40 @@ def update_config_with_suggestion(self, config, suggestion):
             assignments.pop("max_lr", None)
 
         config.update(assignments)
+
+
+class SigOptDynamicImagenetExperiment(SigOptImagenetExperiment):
```
subutai: Best to move this out of this file.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -338,16 +338,29 @@ def _process_config(self, config):
...
             self.suggestion = self.sigopt.get_next_suggestion()
             self.sigopt.update_config_with_suggestion(config, self.suggestion)
             print("SigOpt suggestion: ", self.suggestion)
             print("Config after Sigopt:")
             pprint(config)
             self.epochs = config["epochs"]
 
+            # Get names of performance metrics.
+            assert "metrics" in config["sigopt_config"]
+            self.metric_names = [metric["name"] for metric in config["sigopt_config"]["metrics"]]
```
subutai: Should we assert that at least one of the metrics is called 'mean_accuracy'?

mvacaporale: Not necessarily. It's hard to think of a case where one wouldn't use 'mean_accuracy', ut the user has no requirement to do so

subutai: But if they don't use mean_accuracy, the if statement in _process_result() would prevent the observation from being recorded:

`if result["mean_accuracy"] > 0.0:`

nupic/research/frameworks/sigopt/sigopt_experiment.py
mvacaporale: Made this change.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: Good point. Didn't think of that. See
https://github.com/numenta/nupic.research/pull/217

PR #231, opened 2020-04-07
Updates for Mish (and another small thing)

mvacaporale:
This PR includes function and module classes for the Mish activation (nothing involving kwinners just yet).

As well, it includes a new `ray_custom_loggers_2.py` as a potential candidate to replace the old loggers. The new functionality enables it to work with `tensorflow>=2.0.0`. I've kept these modifications separate for now as I'm still testing them. (@lucasosouza Please try them out and tell me what you think. Once again, nothing is new, other than the tensorflow version that's being used.)


nupic/research/frameworks/pytorch/functions/mish.py
```diff
@@ -0,0 +1,34 @@
...
+
+def mish(x):
+    """
+    Applies the mish function element-wise:
+    mish(x) = x * tanh(softplus(x)) = x * tanh(ln(1 + exp(x)))
+
+    Code adapted from:
+       https://github.com/digantamisra98/Mish/blob/master/Mish/Torch/functional.py
+    """
+    return x * torch.tanh(F.softplus(x))
```
lucasosouza, line 34: Is there an option to do in_place as relu is done, since it is not dependent on any variable? We found a significant increasing performance switching to inplace=True in relu, wondering if we will lose this switching to an inhouse implementation.

For reference, the C implementation of PyTorch activation functions is here: https://github.com/pytorch/pytorch/blob/34b32ca91491147cca2418b955141944d4492bec/aten/src/ATen/native/Activation.cpp

mvacaporale: This is a good point. Unfortunately, mish cannot be done in place. Inspecting the function `x * tanh(softplus(x))` shows you need at least two copies of the network. I may edit this in future iterations to use as little GPU memory as possible.

PR #256, opened 2020-05-14
RES-1481: Refactor ImagenetExperiment to maintains the current epoch number internally

lscheinkman:
In order to make the checkpoints self contained we need to keep the current epoch as part of the `ImagenetExperiment` instead of ray tune or any other external framework

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -35,7 +35,6 @@ dependencies:
   - torchvision=0.5.0
   - tqdm
   - h5py
-  - grpcio>=1.24.3. # needed for tensorboard
 
   # ray[debug,dashboard]
   - aiohttp
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
@@ -415,8 +415,8 @@ class ImagenetExperiment:
                 current_batch *= self.train_loader.sampler.num_replicas
             self.logger.debug("End of batch for rank: %s. Epoch: %s, Batch: %s/%s, "
                               "loss: %s, Learning rate: %s num_images: %s",
-                              self.rank, self.current_epoch, current_batch, total_batches, loss,
-                              self.get_lr(), num_images)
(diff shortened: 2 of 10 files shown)


lscheinkman:
> I think we also need something in imagenet_experiment's run_epoch that tells imagenet_tune to stop based on the epoch. We should put a stop criterion that's based on self.epoch==self.epochs, and then that should be specified in the stop_criterion in imagenet_tune.

Correct, but that will be in a separate PR. See RES-1480

subutai:
> > I think we also need something in imagenet_experiment's run_epoch that tells imagenet_tune to stop based on the epoch. We should put a stop criterion that's based on self.epoch==self.epochs, and then that should be specified in the stop_criterion in imagenet_tune.
> 
> Correct, but that will be in a separate PR. See RES-1480

Ok, got it.

subutai:
> One small comment. Should we make self.epoch to be self.current_epoch? Confusing with self.epoch and self.epochs.

Alternatively we could rename self.epochs

lscheinkman:
> > One small comment. Should we make self.epoch to be self.current_epoch? Confusing with self.epoch and self.epochs.
> 
> Alternatively we could rename self.epochs

Ray tune and some of the mixins refers to the `epoch` as `iteration`.

What do you prefer? `self.epochs => self.num_epochs` or `self.epoch => self.current_epoch` or `self.epoch => self.iteration` ?

subutai:
> > > One small comment. Should we make self.epoch to be self.current_epoch? Confusing with self.epoch and self.epochs.
> > 
> > 
> > Alternatively we could rename self.epochs
> 
> Ray tune and some of the mixins refers to the `epoch` as `iteration`.
> 
> What do you prefer? `self.epochs => self.num_epochs` or `self.epoch => self.current_epoch` or `self.epoch => self.iteration` ?

I don't like iteration for ImagnetExperiment - iteration is too general (which makes sense for Ray). Epoch is more specific to training in networks and widely used in ML.  I'm fine with either of your first two suggestions! Maybe the first one is easier and smaller changes all around.

PR #261, opened 2020-05-22
Refactor create_train_dataloader, move some of it into a classmethod

mrcslws:
This change factors a function `create_train_dataset` out of `create_train_dataloader`, and it puts `create_train_dataloader` functionality inside a method on the ImagenetExpreriment. The method is a classmethod so that it can be easily used by analysis tools while also being overrideable by experiment classes.

This change also does this for `create_validation_dataset`.

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/experiment_utils.py
+++ b/nupic/research/frameworks/pytorch/imagenet/experiment_utils.py
@@ -23,10 +23,8 @@ import socket
 from contextlib import closing
 
 import h5py
-import torch
 from torch.nn.modules.batchnorm import _BatchNorm
 from torch.optim.lr_scheduler import OneCycleLR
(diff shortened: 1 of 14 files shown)


mrcslws:
@lscheinkman @subutai 

With this change, in analysis tools we can create an experiment's validation loader by calling `config["experiment_class"].create_validation_dataloader(config, device)`.

In a future PR I plan to make this same change for `ImagenetExperiment.create_model`. It will become a class method that is ~identical to the current `create_model_from_config`. When we want to use an analysis script, rather than calling `create_model_from_config(config, device)`, we'll call `config["experiment_class"].create_model(config, device)`.

lscheinkman:
@mrcslws Nice!
This is exactly what @lucasosouza needed for https://github.com/numenta/nupic.research/pull/257

PR #267, opened 2020-05-28
Make aggregation of distributed results extensible

mrcslws:
Enables mixins to handle aggregation.

In a future PR I'm adding a mixin that saves the training loss for every batch and stores it as a list in the result dict. The current PR allows me to aggregate those lists.

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
@@ -66,6 +66,8 @@ class ImagenetTrainable(Trainable):
         return resource
 
     def _setup(self, config):
+        self.experiment_class = config["experiment_class"]
+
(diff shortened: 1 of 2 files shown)


nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -195,13 +192,13 @@ def _create_workers(self, config):
 
         self._process_config(config)
 
-        experiment_class = config["experiment_class"]
+        self.experiment_class = config["experiment_class"]
```
lscheinkman, line 195: I would initialize this field at the beginning of `_setup` method instead `_create_workers` method

mrcslws: Good point.

PR #272, opened 2020-05-29
RES-1483: Make sure checkpoint used to restore does not override model initial checkpoint state

lscheinkman:


nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -83,24 +83,27 @@ def _setup(self, config):
...
-        if self._checkpoint_file is not None:
-            with open(self._checkpoint_file, mode="rb") as f:
+        self._restored = False
+        if self.restore_checkpoint_file is not None:
+            with open(self.restore_checkpoint_file, mode="rb") as f:
                 state = pickle.load(f)
                 self._restore(state)
                 self._restored = True
 
+        elif config.get("checkpoint_at_init", False):
```
subutai, line 103: For consistency, wouldn't we still want to save the checkpoint at init if that flag is set to True?

lscheinkman: Are you sure? It would just be a copy of the original checkpoint?

lscheinkman:
> How would we run the _continue experiments? Would we specify the checkpoint on the command line?

`python run.py -c checkpoint` to resume from a specific checkpoint file or `python run.py --restore` to continue training from the last known checkpoint.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
subutai: It would, just wondering for consistency. I'm fine either way.

PR #274, opened 2020-06-01
Two changes: No more ComplexLoss. Explicit regularization loss.

mrcslws:
This change adds two extensibility hooks to `train_model` and `evaluate_model`.

The first one, `to_device_fn`, makes it possible to do Knowledge Distillation and Maxup using `train_model`. So we no longer need the `ComplexLoss` mixin.

The second, `complexity_loss_fn`, makes it much less messy to write a correct `loss_function`. Now the loss is split into `error_loss` and `complexity_loss`. The error loss depends on the data and needs to know how to handle "reduction". The complexity loss depends only on the model and hence doesn't need to perform reductions over the batch.

Note that "ComplexLoss" and "complexity_loss" have nothing to do with each other. Coincidentally similar words.


nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -379,8 +379,10 @@ def validate(self, loader=None):
                 model=self.model,
                 loader=loader,
                 device=self.device,
-                criterion=self.loss_function,
+                criterion=self.error_loss,
+                complexity_loss_fn=self.complexity_loss,
                 batches_in_epoch=self.batches_in_epoch,
+                to_device_fn=self.send_data_to_device,
```
lscheinkman, line 385: `to_device_fn` name is not clear to me. The name implies it is only used to transform from one device to another.
But from what I understood this callback is also used to apply final transformations to the model results, such as transforming the result from "one hot" results to some other complex results, not necessary to another device. Maybe we could call it `final_transform` or `result_transform`

nupic/research/frameworks/pytorch/model_utils.py
```diff
@@ -210,12 +243,21 @@ def evaluate_model(
         loader.close()
 
     correct = correct.item()
+
+    if total > 0:
+        loss /= total
+
+    if complexity_loss_fn is not None:
```
mrcslws, line 250: I should move this up to be inside the `torch.no_grad()` block.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mrcslws: Hmm. One consideration: it is responsible for moving the data and target to the device. This is necessary because in some cases it chooses not to move the `target`, e.g. with Knowledge Distillation. Simply calling it a `final_transform` doesn't capture that requirement.

mrcslws: `transform_to_device_fn`?

PR #275, opened 2020-06-01
Res 1488 batch norm improvements

lucasosouza:
Adds option to configure batch norm momentum.

It is a small change with a small impact - but it is a new hyperparameter that can be worth tuning. Maybe we should move this hyperparameter to ImageNet Experiment as opposed to a mixin? 

Update: Edited to be a parameter of the network as suggested


mrcslws:
A more natural place for this code might be in the model-creation code in `resnets.py`. Create the `nn.BatchNorm2d` with the correct momentum in the first place, rather than creating and overriding later.

PR #278, opened 2020-06-02
Properly print distributed validation results

mrcslws:


nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -109,9 +111,23 @@ def _setup(self, config):
...
+                    status = []
+                    for w in self.procs:
+                        status.append(w.validate.remote())
+
+                    if ray_utils.check_for_failure(status):
+                        results = ray.get(status)
+                        ret = self.experiment_class.aggregate_validation_results(
+                            results)
+                        self.logger.info(ret)
+
```
lscheinkman, line 128: I would move this code to after the `_should_stop` check to avoid duplicate results during the restore

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -109,9 +111,23 @@ def _setup(self, config):
     def _train(self):
         self.logger.debug(f"_train: {self._trial_info.trial_name}({self.iteration})")
         try:
-            # Check if restore checkpoint file fulfills the stop criteria on first run
             if self._first_run:
                 self._first_run = False
+
+                if self._validate_immediately:
```
lscheinkman, line 117: This is not necessary before the first epoch as it was originally intended. 
During the restore the `first_run` is the last epoch stored with the checkpoint

PR #287, opened 2020-06-07
RES-1492: Add script to run distributed imagenet experiments on a single node multi GPU without ray tune

lscheinkman:
This PR removes ray dependencies from `imagenet_experiment.py` and adds a new script script to run distributed imagenet experiments on a single node using multi GPUs without ray tune.
The new script is called `local_run.py` and uses the same command line interface as the original ray tune version. For example:
```
python local_run.py -e default_sparse_1000
```
Will run the `default_sparse_1000` experiment on the local node using all available GPUs

### DIFF
--- a/projects/imagenet/local_run.py
+++ b/projects/imagenet/local_run.py
@@ -31,12 +31,12 @@ from functools import partial
 import torch
 import torch.multiprocessing as multiprocessing
 
-from nupic.research.frameworks.pytorch.imagenet import imagenet_run
 from nupic.research.frameworks.pytorch.imagenet import (
-    mixins, ImagenetExperiment
+    ImagenetExperiment,
+    imagenet_run,
+    mixins,
 )
 from nupic.research.frameworks.sigopt.sigopt_experiment import SigOptImagenetExperiment
-
 
 multiprocessing.set_start_method("spawn", force=True)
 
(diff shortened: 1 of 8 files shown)


projects/imagenet/local_run.py
```diff
@@ -0,0 +1,188 @@
...
+def create_trials(config):
+    """
+    Create trial configuration for each trial variant evaluating 'ray.tune'
+    functions (grid_search, sample_from, ...) into its final values and
+    creating the local and log dir for each trial
+
+    :param config: Ray tune configuration with 'ray.tune' functions
+    :return: list of dict for each trial configuration variant
+    """
+    from nupic.research.support.ray_utils import generate_trial_variants
```
subutai, line 63: Do we need all this? Seems like overkill and replicating ray tune. All we do is use tune to get a random seed. Why don't we just move away from using ray tune altogether in our regression tests? We could just use regular random() to figure out the seed.

lscheinkman: We will not use this script for regression tests. The regression tests will use the `imagenet_run.run(config,logger=None, on_checkpoint=None)` API from the unit test using a custom logger to validate the results without any ray or ray.tune dependencies.

This script is just a direct replacement to `run.py` to serve as example.

PR #301, opened 2020-06-20
Kd improvements

lucasosouza:
Some improvements to Knowledge Distillation:
- Option to pass list of models instead of single model for ensembles. Can weight the models
- Adds a temperature factor that can be linearly decayed

Also includes a CompositeLoss mixin, similar to ComplexLoss mixin that was removed. It is used for fast prototyping of different features. Renamed it to avoid confusion with complexity loss.


nupic/research/frameworks/pytorch/imagenet/mixins/composite_loss.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Defines a new training loop that has direct access to Experiment attributes
+    and can be customized for more complex loss functions
+    """
+
+    def train_epoch(self):
+        """Overwrites train model to use private train_model method"""
+        self._train_model()
+
+    def _train_model(self):
```
mrcslws, line 38: You could just put this code in the `train_epoch` body. Is there a reason for `_train_model` to exist? This makes it seem like you're adding two hooks, when fundamentally you're only adding one (`calculate_composite_loss`).

nupic/research/frameworks/pytorch/imagenet/mixins/knowledge_distillation.py
```diff
@@ -133,6 +201,66 @@ def get_execution_order(cls):
         return eo
 
 
+class KnowledgeDistillationCL(KnowledgeDistillation):
```
mrcslws, line 204: Does "CL" stand for "combined loss"? When I first saw this, my first guess was that it meant `CompositeLoss` and was indicating that it needs to be used with `CompositeLoss`.

nupic/research/frameworks/pytorch/imagenet/mixins/composite_loss.py
lucasosouza: That is a good point, it made sense before but not now. I will merge now but I will make some changes to that class in the next PR.

nupic/research/frameworks/pytorch/imagenet/mixins/knowledge_distillation.py
lucasosouza: It does. I initially used CompositeLoss mixin, because I needed all memory possible and I was worried of leaving all the soft targets in memory (which is the only way to do it with the existing approach of transform data and target + separate loss function). However, upon realizing the targets have a very small memory footprint, even when one they are soft targets and not just indices, I switched to the current approach.

lucasosouza: The name composite loss is due to the loss function being composed of of several partial loss functions in this approach. Might not be a good term after all, I just ran out of creativity =)

PR #306, opened 2020-06-24
RES-1509 Cutmix

lucasosouza:
Implements Cutmix regularization as a mixin. Includes a mixin that combines with simple forms with Knowledge Distillation.

Cutmix is considered to be a state of the art algorithm in regularization. It combines two ideas:
* Mixup: mixes two classes by mixing both the input and the target. Mixup linearly combines the inputs, and can distort the image to unrealistic settings.
* Cutout: Also known as regional dropout, replaces patches of the image by either black patches or patches with random noise. Replacing a patch of the image incurs in loss of information.
* Cutmix: combine both methods, by replacing random patches in the image with patches from other images in the dataset, and combining the target labels accordingly.

Improvements seen in 100 classes experiments - at 200 epochs, acc up from 87.16 to 87.32. Tests in the 1000 classes, either regular training or with knowledge distillation, did not result in improvements.


karangrewal:
It all looks fine to me. Since I'm also less familiar with this code base, let's see what Subutai says.

PR #324, opened 2020-07-27
Res 1602

karangrewal:
Added unit tests to to ensure that target variables are being transformed properly according to the continual learning setup found in the [benchmarks repository](https://github.com/GMvandeVen/continual-learning). All unit tests pass.

Also, 2 other small changes to previous files:
1. fixed a small error in `train_model.py` to make sure it selects the number of output nodes in an MLP classifier based on the scenario/setup.
2. in `subdataset.py`, set `download=True` as the default argument to the get_datasets function, as the unit tests fail otherwise


nupic/research/frameworks/continuous_learning/multihead/subdataset.py
```diff
@@ -36,7 +36,7 @@
 from torchvision.datasets import MNIST
```
lscheinkman, line 38: How is this class different from [split_dataset](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/dataset_utils/dataset_utils.py#L168) defined in `dataset_utils.py`? It looks like this file could be replaced by a few functions in dataset_utils.py

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+from torch.utils.data import DataLoader
+
+from nupic.research.frameworks.continuous_learning.multihead.subdataset import (
+    get_datasets,
+)
+
+
+def setup(dataset_name, scenario):
+    train_datasets, test_datasets = get_datasets(dataset_name, scenario=scenario)
```
lscheinkman, line 38: This test requires downloading MNIST dataset to run which could take a long time. Unit tests are supposed to be fast and only check for specific functionality.
If you need a dataset to test you should use [FakeDataLoader](https://github.com/numenta/nupic.research/blob/53c19f5eec2da8091219f270f124c5aef74bbe02/nupic/research/frameworks/pytorch/test_utils/fake_data.py#L30) instead

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,95 @@
...
+    return train_loaders, test_loaders
+
+
+class MultiHeadTest(unittest.TestCase):
+
+    def test_task_scenario(self):
+        """
+        Test all target values to be in {0, 1} in the task scenario.
+        """
+        train_loaders, test_loaders = setup("splitMNIST", "task")
```
lscheinkman, line 52: You may need to separate the logic that loads the dataset (`dataset_class`) from the `scenario` functionality to be able to unit test only the `scenario`. Maybe pass `dataset_class` as parameter to `get_datasets`

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,95 @@
...
+        for test_loader in test_loaders:
+            for _data, target in test_loader:
+                self.assertLess(target.max().item(), 2)
+                self.assertGreaterEqual(target.min().item(), 0)
+
+    def test_domain_scenario(self):
+        """
+        Test all target values to be in {0, 1} in the domain scenario.
+        """
+        train_loaders, test_loaders = setup("splitMNIST", "domain")
```
lscheinkman, line 66: Same comment as `test_task_scenario`

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,95 @@
...
+            for _data, target in test_loader:
+                self.assertLess(target.max().item(), 2)
+                self.assertGreaterEqual(target.min().item(), 0)
+
+    def test_class_scenario(self):
+        """
+        Test all target values to be changing based on the task number in the class
+        scenario.
+        """
+        train_loaders, test_loaders = setup("splitMNIST", "class")
```
lscheinkman, line 81: Same comment as test_task_scenario

nupic/research/frameworks/continuous_learning/multihead/subdataset.py
```diff
@@ -36,7 +36,7 @@
...
 
 
-def get_datasets(dataset_name, scenario, download=False):
+def get_datasets(dataset_name, scenario, download=True):
```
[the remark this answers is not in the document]
karangrewal: I originally changed this since the code would break since the unit tests required the data to run, but since we're going to use a FakeDataLoader, I will change it back to `download=False`.

nupic/research/frameworks/continuous_learning/multihead/subdataset.py
karangrewal: [split_dataset](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/dataset_utils/dataset_utils.py#L168) doesn't allow for applying a target_transform, so I need to figure out how to be able to transform target variables using just dataset_utils.py

karangrewal: I think I resolved the issue, but need to make a slight modification to [split_dataset](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/dataset_utils/dataset_utils.py#L168) to include a target_transform

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Since the target transforms interface with the data class directly, I think I'll need to deal with [FakeData](https://pytorch.org/docs/stable/torchvision/datasets.html#fakedata) instead of the data loader.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Done.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Done.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Done.

nupic/research/frameworks/pytorch/dataset_utils/dataset_utils.py
```diff
@@ -189,7 +191,13 @@ def split_dataset(dataset, groupby):
     _, indices = list(
         zip(*(sorted(list(indices_by_group.items()), key=lambda x: x[0])))
     )
-    return [Subset(dataset, indices=i) for i in indices]
+    subsets = [Subset(dataset, indices=i) for i in indices]
+
+    # Apply target_transform, if defined
+    if target_transform is not None:
```
lscheinkman, line 197: Why not apply the transforms to the dataset targets before splitting? 
Pytorch's torchvision datasets already support that, there is no need to do it only when spliting.
See `target_transform` argument in https://pytorch.org/docs/stable/torchvision/datasets.html#mnist

nupic/research/frameworks/pytorch/test_utils/fake_data.py
```diff
@@ -18,15 +18,50 @@
...
 from torchvision.datasets import FakeData
-from torchvision.transforms import ToTensor
 
 __all__ = [
+    "FakeDataPredefinedTargets",
     "FakeDataLoader",
 ]
 
 
+class FakeDataPredefinedTargets(FakeData):
```
lscheinkman, line 32: Why is this necessary? Why do you need predefined `targets` for fake data?

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -28,18 +28,38 @@
...
 )
+from nupic.research.frameworks.pytorch.dataset_utils.dataset_utils import split_dataset
+from nupic.research.frameworks.pytorch.test_utils.fake_data import (
+    FakeDataPredefinedTargets,
+)
+
+
+def setup(scenario):
+    fake_train_dataset = FakeDataPredefinedTargets(image_size=(1, 28, 28))
+    fake_test_dataset = FakeDataPredefinedTargets(image_size=(1, 28, 28))
```
lscheinkman, line 44: Why not just `FakeData`?

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -28,18 +28,38 @@
...
+
+    if scenario in ("task", "domain"):
+        target_transform = transforms.Lambda(lambda y: y % 2)
+    else:
+        target_transform = None
 
+    train_datasets = split_dataset(fake_train_dataset,
+                                   groupby=lambda x: x[1] // 2,
+                                   target_transform=target_transform)
```
lscheinkman, line 53: `target_transform` should be applied to the dataset before splitting

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -28,18 +28,38 @@
...
+    test_datasets = split_dataset(fake_test_dataset,
+                                  groupby=lambda x: x[1] // 2,
+                                  target_transform=target_transform)
```
lscheinkman, line 56: target_transform should be applied to the dataset before splitting

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -28,18 +28,38 @@
...
 
-def setup(dataset_name, scenario):
-    train_datasets, test_datasets = get_datasets(dataset_name, scenario=scenario)
     train_loaders = [DataLoader(train_dataset, batch_size=256, shuffle=True)
                      for train_dataset in train_datasets]
-    test_loaders = [DataLoader(test_dataset, batch_size=1000, shuffle=True)
+    test_loaders = [DataLoader(test_dataset, batch_size=500, shuffle=True)
```
lscheinkman, line 60: Should use smaller batch sizes for unit tests. Unit test should not care about accuracy, the purpose of batch size in here should be the minimum batch size possible to cover the functionality being tested. 500 seems to large for that purpose.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -78,18 +98,45 @@ def test_class_scenario(self):
...
         for task_num, train_loader in enumerate(train_loaders):
             min_target, max_target = 2 * task_num, 2 * task_num + 1
             for _data, target in train_loader:
-                self.assertLess(target.max().item(), max_target + 1)
-                self.assertGreaterEqual(target.min().item(), min_target)
+                # print(target, min_target, max_target)
+                try:
+                    self.assertLess(target.max().item(), max_target + 1)
+                    self.assertGreaterEqual(target.min().item(), min_target)
+                except AssertionError:
```
lscheinkman, line 109: Why are you ignoring the `AssertionError`?

nupic/research/frameworks/pytorch/test_utils/fake_data.py
karangrewal: because the `split_dataset` function assumes the dataset class has a `targets attribute` (which MNIST does), but the `FakeData` class does not.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: (same comment as above)

because the `split_dataset` function assumes the dataset class has a `targets` attribute (which MNIST does), but the `FakeData` class does not.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: ok, that makes sense.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Ok, I'll change to 50 for tests.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: oops, I wrote this while debugging, I will fix it now.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
lscheinkman: Maybe the fix should be in the `split_dataset` function instead

nupic/research/frameworks/pytorch/test_utils/fake_data.py
lscheinkman: This is a new code to `split_dataset` regarding target_transform. If you apply target_transform before splitting and restore `split_dataset` to its original version, you won't need to access `targets` from ``split_dataset` and will be able to use `FakeData`

nupic/research/frameworks/pytorch/dataset_utils/dataset_utils.py
karangrewal: the `split_dataset` function does the splitting based on the labels, so if we make all labels {0, 1} and then split, it will think there are only 2 classes.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: reverted changes to `split_dataset`, so it no longer uses `target_transform`.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,170 @@
...
+from nupic.research.frameworks.continuous_learning.multihead.multihead import (
+    active_class_outputs,
+    get_active_classes,
+    get_target_transform,
+)
+from nupic.research.frameworks.pytorch.test_utils.fake_data import FakeDataLoader
+from nupic.torch.modules import Flatten
+
+
+def setup(scenario):
```
lscheinkman, line 41: `scenario` arg is not being used, it could be renamed to `setup_dataloaders`

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class MultiHeadTest(unittest.TestCase):
+
+    def test_active_class_outputs_task(self):
+        """
+        Test the multihead output size for the task scenario.
+        """
+        model = setup_model(784, 10)
+        train_loaders, test_loaders = setup("task")
```
lscheinkman, line 63: `scenario` arg is not being used

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,170 @@
...
+        for data, _target in test_loaders[1]:
+            output = active_class_outputs(model, data, [2, 3])
+            self.assertEqual(output.shape[1], 2)
+
+    def test_active_class_outputs_domain(self):
+        """
+        Test the multihead output size for the domain scenario.
+        """
+        model = setup_model(784, 2)
+        train_loaders, test_loaders = setup("domain")
```
lscheinkman, line 76: `scenario` arg is not being used

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
```diff
@@ -0,0 +1,170 @@
...
+        for data, _target in test_loaders[1]:
+            output = active_class_outputs(model, data, None)
+            self.assertEqual(output.shape[1], 2)
+
+    def test_active_class_outputs_class(self):
+        """
+        Test the multihead output size for the class scenario.
+        """
+        model = setup_model(784, 10)
+        train_loaders, test_loaders = setup("task")
```
lscheinkman, line 89: `scenario` arg is not being used

nupic/research/frameworks/pytorch/test_utils/fake_data.py
```diff
@@ -19,8 +19,8 @@
 # ----------------------------------------------------------------------
 
 from torch.utils.data import DataLoader
+from torchvision import transforms
```
lscheinkman, line 22: Not really changing anything in this file, maybe revert to original

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Done.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Removed.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Removed.

tests/unit/frameworks/continuous_learning/multihead/multihead_test.py
karangrewal: Removed.

nupic/research/frameworks/pytorch/test_utils/fake_data.py
karangrewal: need to add blank line at the end of the original file, because it gives a style error otherwise.

see here: https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/test_utils/fake_data.py

nupic/research/frameworks/continuous_learning/multihead/multihead.py
```diff
@@ -103,8 +99,6 @@ def test(model, loader, criterion, device, allowed_classes=None):
     with torch.no_grad():
         for data, target in tqdm(loader, desc="Test", leave=False):
             data, target = data.to(device), target.to(device)
-            # if allowed_classes is not None:
-            #     target = target - allowed_classes[0]
```
lscheinkman, line 107:
```
output = model(data)
if allowed_classes is not None:
    output = output[:, allowed_classes]
```
Should be replaced with 
```   
output = active_class_outputs(model, data, active_classes)
```

PR #325, opened 2020-07-27
LR Range Test Mixin

mvacaporale:
This PR includes a mixin for the LR range test (described here https://arxiv.org/abs/1803.09820)

To aid in the functionality
* There are some changes to `LogEveryLoss` so that the LR can also be logged at every timestep
* There's a new LinearLRScheduler (similar to one-cycle with a linear ramp-up, but with no ramp-down portion)
* The ImagenetExperiment class now has a class method `create_lr_scheduler` which can be overridden.


nupic/research/frameworks/pytorch/imagenet/mixins/log_every_loss.py
```diff
@@ -35,12 +35,14 @@ def setup_experiment(self, config):
         super().setup_experiment(config)
         self.error_loss_history = []
         self.complexity_loss_history = []
+        self.learning_rate_history = []
```
mrcslws, line 38: There is already a LogEveryLearningRate mixin. I think it's better to keep these separate.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,111 @@
...
+from copy import deepcopy
+
+from pandas import DataFrame
+from ray.tune.utils import flatten_dict
+
+from nupic.research.frameworks.pytorch.imagenet import mixins
+from nupic.research.frameworks.pytorch.lr_scheduler import LinearLRScheduler
+
+
+class LRRangeTest(mixins.LogEveryLoss):
```
mrcslws, line 20: Is subclassing necessary? I think it's cleaner to try to keep all of the mixins orthogonal and let experiments choose which ones to opt into.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,111 @@
...
+        # Ensure all epochs get validated.
+        assert "epochs" in config
+        config["epochs_to_validate"] = range(-1, config["epochs"])
+
+        # Init saving of test results.
+        self.save_lr_test_to_csv = config.get("save_lr_test_to_csv", False)
+        if self.save_lr_test_to_csv:
+            assert "logdir" in config, (
+                "The logdir must be specified to save the lt-range test resutls.")
+            self.logdir = config["logdir"]
```
mrcslws, line 59: This already happens in ImagenetExperiment: https://github.com/numenta/nupic.research/blob/57b4d77e062c97c7ad62b0c681bb953442375ee1/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py#L204

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # Save config for later - used to aggregate results.
+        self._config = deepcopy(config)
+        super().setup_experiment(config)
```
mrcslws, line 63: Is there a reason you're calling super() at the end rather than the beginning? The convention is to always do it at the beginning unless you have a good reason not to. (e.g. profiling mixins)

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+        lr_scheduler_args.update(epochs=config["epochs"])
+        lr_scheduler_args.update(steps_per_epoch=total_batches)
+
+        return lr_scheduler_class(optimizer, **lr_scheduler_args)
+
+    def post_batch(self, *args, **kwargs):
+        """Increase lr after every batch."""
+        self.lr_scheduler.step()
+        super().post_batch(*args, **kwargs)
```
mrcslws, line 81: Similar question: super() should probably be at the beginning?

nupic/research/frameworks/pytorch/imagenet/mixins/log_every_loss.py
mvacaporale: Didn't realize, thanks. I'll revert the change.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
mvacaporale: It felt odd not to subclass. The idea of the test is to compare the training loss against the learning rate at every mini-batch step, so it makes sense to make sure they are logged.

I think I have an alternative idea, however. Let me know what you think of the `create_lr_test_experiment` function in the new changes.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
mvacaporale: Good point. Fixed.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
mvacaporale: Fair point.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
mvacaporale: Fair point. I've fixed both places where this occurs.

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,99 @@
...
+        """Increase lr after every batch."""
+        super().post_batch(*args, **kwargs)
+        self.lr_scheduler.step()
+
+    @classmethod
+    def get_execution_order(cls):
+        eo = super().get_execution_order()
+        eo["setup_experiment"].insert(0, cls.__name__ + ": initialize")
+        eo["create_lr_scheduler"] = cls.__name__ + ": create_lr_scheduler"
+        eo["post_batch"].insert(0, cls.__name__ + ": linearly increase lr")
```
mrcslws, line 78: Remember to update these, now that super() is called at the beginning. (Append to the end rather than inserting at beginning)

nupic/research/frameworks/pytorch/imagenet/mixins/lr_range_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+        lr_scheduler_args.update(epochs=config["epochs"])
+        lr_scheduler_args.update(steps_per_epoch=total_batches)
+
+        return lr_scheduler_class(optimizer, **lr_scheduler_args)
+
+    def post_batch(self, *args, **kwargs):
+        """Increase lr after every batch."""
+        super().post_batch(*args, **kwargs)
+        self.lr_scheduler.step()
```
mrcslws, line 71: This is going to step twice at the end of each epoch, since ImagenetExperiment will step again post_epoch. We should probably make this logic in the ImagenetExperiment extensible. Probably the simplest and most reusable way: we could add a param like `lr_scheduler_step_every_batch` as an override to the config, have it default to None. If it's True or False, the class obeys it. If the override isn't set, ImagenetExperiment can set it to `isinstance(lr_scheduler, (OneCycleLR, ComposedLRScheduler))`. So you'd want to edit ImagenetExperiment.post_batch and post_epoch to obey this bool.

Then the mixin's post_batch won't be necessary, you can just set the "lr_scheduler_step_every_batch" parameter.

mvacaporale: Good idea. I made this change.

Note: I realized in setting `lr_scheduler_step_every_batch` from the `setup_experiment` of the `LRRangeTest`, that I needed this done before super's setup, so I changed the order back.

PR #327, opened 2020-07-30
Change name of `...shortcut` to `...shortcut.conv`.

mvacaporale:


mrcslws:
For any onlookers, note that the name of the module isn't being changed. Just the name we pass into a callback.

PR #328, opened 2020-07-31
Workaround OneCycleLR state_dict bug

mrcslws:
Workaround for https://github.com/pytorch/pytorch/issues/42376. Avoids running into this issue in future checkpoints. Adds a script for repairing existing checkpoints.

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
@@ -719,7 +719,12 @@ class ImagenetExperiment:
 
         if self.lr_scheduler is not None:
             with io.BytesIO() as buffer:
-                serialize_state_dict(buffer, self.lr_scheduler.state_dict())
+                state_dict = self.lr_scheduler.state_dict()
+                if "anneal_func" in state_dict:
+                    # FIXME: This is a workaround for a PyTorch bug.
+                    # https://github.com/pytorch/pytorch/issues/42376
+                    del state_dict["anneal_func"]
+                serialize_state_dict(buffer, state_dict)
                 state["lr_scheduler"] = buffer.getvalue()
 
         if self.mixed_precision:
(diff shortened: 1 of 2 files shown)


nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -719,7 +719,12 @@ def get_state(self):
 
         if self.lr_scheduler is not None:
             with io.BytesIO() as buffer:
-                serialize_state_dict(buffer, self.lr_scheduler.state_dict())
+                state_dict = self.lr_scheduler.state_dict()
+                if "anneal_func" in state_dict:
+                    # FIXME: This is a workaround for a PyTorch bug.
+                    # https://github.com/pytorch/pytorch/issues/42376
+                    del state_dict["anneal_func"]
```
lscheinkman, line 726: You should store the `anneal_strategy` instead and restore the correct function on load.

projects/imagenet/repair_onecyclelr_checkpoint.py
```diff
@@ -0,0 +1,69 @@
...
+
+    fix_needed = False
+    if "lr_scheduler" in checkpoint:
+        print("Loading LR scheduler state dict (this might take a few minutes)")
+        with io.BytesIO(checkpoint["lr_scheduler"]) as buf:
+            lr_sched_state_dict = deserialize_state_dict(buf)
+
+        if "anneal_func" in lr_sched_state_dict:
+            fix_needed = True
+            del lr_sched_state_dict["anneal_func"]
```
lscheinkman, line 49: You should store the `anneal_strategy` instead and restore the correct function on load.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mrcslws: I think that would be overkill. The state dict is used for time-varying attributes. There is no supported way to change the anneal strategy after the LR scheduler is instantiated. They include it in the state dict because they're lazily saving the whole `self.__dict__` rather than writing proper `state_dict` methods.

projects/imagenet/repair_onecyclelr_checkpoint.py
mrcslws: I think this would be overkill. (see above)

lscheinkman: makes sense

PR #331, opened 2020-08-12
RES 1613 - Quantization Aware Training

lucasosouza:
Working version of QAT. I made some simplifications to make it easier to merge with existing code, I might reintroduce in the next PR. See related PR in nupic.hardware for all changes.

Create_model was divided into three functions:
- create_model: create the network
- restore_checkpoint: restore checkpoint 
- transform_model: default just moves to device.

Create_model can still perform both of the other functions, in order to maintain backwards compatibility, but that is no longer the default behavior in ImageNet.

The reasoning behind this change is when training QAT from a checkpoint, the model has to be 1. created, 2. have the checkpoint restored, and then 3. transformed. When re-loading a model trained with QAT, the order is changed: we first have to 1. create the model, 2. transform and then 3. restore the checkpoint, since the checkpoint from the finetuned QAT model will have all the extra parameters. So the flexibility to be able to change the order of these operations was necessary, and the added flexibility will also be useful moving forward.


nupic/research/frameworks/pytorch/imagenet/mixins/rezero_weights.py
```diff
@@ -44,8 +44,8 @@ def setup_experiment(self, config):
                               float(nonzero_params_sparse2) / params_sparse)
 
     @classmethod
-    def create_model(cls, config, device):
-        model = super().create_model(config, device)
+    def create_model(cls, config):
```
lscheinkman, line 47: refactoring this method signature will break any code that rely on the device being passed to `create_model`.
Just run `git grep 'create_model'` in this and any other repo that depends on `nupic.research` to get a list of files that need to be updated as well.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -366,8 +383,8 @@ def create_model(cls, config, device):
             model_args=config.get("model_args", {}),
             init_batch_norm=config.get("init_batch_norm", False),
             device=device,
-            checkpoint_file=config.get("checkpoint_file", None),
-            load_checkpoint_args=config.get("load_checkpoint_args", {}),
+            checkpoint_file=checkpoint_file,
+            load_checkpoint_args=load_checkpoint_args,
```
lscheinkman, line 387: Why do you need to refactor this method passing `checkpoint_file`, `load_checkpoint_args` as arguments ignoring the values in the `config` dictionary. 
This will break resume and any other script that rely on the `config` to pass the checkpoint file.
This class method is called directly by other scripts without instantiating the experiment. Such as regression and other tests

nupic/research/frameworks/quantization/qat_kwinners.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def forward(self, x):
+        if self.n == 0:
+            self.n = np.prod(x.shape[1:])
+            if not self.local:
+                self.k = int(round(self.n * self.percent_on))
+                self.k_inference = int(round(self.n * self.percent_on_inference))
+
+        if self.training:
+            x = self.kwinner_function(x, self.duty_cycle, self.k, self.boost_strength)
```
lscheinkman, line 128: You should get the latest `nupic.torch`. 
`kwinner_function` was refactored out. see https://github.com/numenta/nupic.torch/pull/74/

nupic/research/frameworks/pytorch/imagenet/mixins/quantization_aware.py
```diff
@@ -0,0 +1,201 @@
...
+        qconfig = QConfig(
+            activation=FakeQuantize.with_args(observer=MovingAverageMinMaxObserver,
+                                              quant_min=0,
+                                              quant_max=255,
+                                              reduce_range=True),
+            weight=default_weight_fake_quant
+        )
+    model.qconfig = qconfig
+
+    # equivalent to quantize.prepare, inplace. require for custom white list
```
lscheinkman, line 146: This was fixed in pytorch 1.6. You can now pass custom white list to `torch.prepare`

nupic/research/frameworks/pytorch/imagenet/network_utils.py
```diff
@@ -92,12 +92,19 @@ def create_model(model_class, model_args, init_batch_norm, device,
     model = model_class(**model_args)
     if init_batch_norm:
         init_resnet50_batch_norm(model)
-    model.to(device)
 
-    # Load model parameters from checkpoint
+    # backwards compatibility
```
lscheinkman, line 96: This comment is confusing. All you doing is checking if the optional `device` argument is `None`. It is confusing to mark that as 'backward compatibility'

nupic/research/frameworks/quantization/qat_kwinners.py
lucasosouza: Do you know if the changes are paramount to make it work? This is working fine on my end, and I pulled nupic.torch less than 2 weeks ago. I wouldn't like to refactor at this PR unless necessary. If it is style or performance reasons I can make the changes in the next PR.

lscheinkman: Did you get the latest nupic.torch from master? It does not work with the latest nupic.torch

nupic/research/frameworks/pytorch/imagenet/network_utils.py
lucasosouza: It is sending the model to device here, if device is not None. That is to maintain backward compatibility in case you only call `create_model` function directly and you expect it to send the model to device. In the new version, the model is only sent to device in the `transform_model` function. Do you propose a different approach?

lscheinkman: No, just remove the comment. I was trying to understand "backwards compatibility" comment from the code without looking at the diffs and it was not clear to me

nupic/research/frameworks/quantization/qat_kwinners.py
lscheinkman: There is no `kwinner_function` in nupic.torch code
https://github.com/numenta/nupic.torch/search?q=kwinner_function&type=Code

nupic/research/frameworks/pytorch/imagenet/mixins/quantization_aware.py
lucasosouza: Not `prepare_qat`, they only updated `prepare` in 1.6

nupic/research/frameworks/pytorch/imagenet/mixins/rezero_weights.py
lucasosouza: Fixed

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lucasosouza: I kept the same signature and added an extra flag whether or not to use those arguments from config

nupic/research/frameworks/quantization/qat_kwinners.py
lucasosouza: Fixed.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -230,7 +231,16 @@ def setup_experiment(self, config):
...
-        self.model = self.create_model(config, self.device)
+        self.model = self.create_model(config, from_checkpoint=False)
+
+        checkpoint_file = config.get("checkpoint_file", None)
+        if checkpoint_file:
+            load_checkpoint_args = config.get("load_checkpoint_args", {})
+            self.restore_checkpoint(self.model, checkpoint_file, load_checkpoint_args)
+
+        # Send to device after loading checkpoint
+        self.transform_model()
```
mrcslws, line 242: This is forcing us to always restore checkpoints on CPU rather than on CUDA, then to move it back to CUDA (even in our main experiments when we're not using QAT yet). Maybe instead, the QuantizationAware mixin should override create_model and put all of this logic there. I'm not sure we need `transform_model` and `restore_checkpoint`, it seems the current ImagenetExperiment is already sufficient to implement this.

Then you could use the `device=None` option in `network_utils.create_model`, but you won't need the `from_checkpoint` bool any more.

lucasosouza: Do you see a big impact in performance? I understand your point, but the added flexibility seemed like a good trade compared to at most a few seconds more only in initialization, but I might have estimated it wrongly.

lucasosouza: And functionally, from a SRP principle, I think restore checkpoints and creating models should be separate methods.

lucasosouza: The last commit should address your concerns. There is still a transform model, but create_model is unchanged, so shouldn't impact existing experiments. I had to a little bit of hacking in the `quantize_qat` script to account for the different order of checkpoint restoration - in a later refactor we should disentangle restore_checkpoint from create model.

PR #333, opened 2020-08-14
Initial Refactor for Experiment Trainables

mvacaporale:
This PR gets the ball rolling on making our experiment classes more extensible. 

It starts with the Trainables:
* Refactoring the initial-validation functionality into a `pre_experiment` method.
* Introducing a `BaseTrainable` and `SupervisedTrainable`

@lscheinkman You'll notice that the notion of an iteration has been kept only with the Trainables and is not imposed on the ImagenetExperiment class.

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
@@ -47,7 +47,7 @@ from nupic.research.support.ray_utils import (
 os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
 
 
-class ImagenetTrainable(Trainable):
+class SupervisedTrainable(Trainable):
     """
     Trainable class used to train resnet50 on Imagenet dataset using ray
     """
@@ -297,7 +297,7 @@ class ImagenetTrainable(Trainable):
         pass
 
 
-class SigOptImagenetTrainable(ImagenetTrainable):
+class SigOptImagenetTrainable(SupervisedTrainable):
     """
     This class updates the config using SigOpt before the models and workers are
     instantiated, and updates the result using SigOpt once training completes.
@@ -379,9 +379,9 @@ def run(config):
         kwargs = dict(zip(kwargs_names, [SigOptImagenetTrainable,
                                          *tune.run.__defaults__]))
     else:
-        imagenet_trainable = config.get("imagenet_trainable", ImagenetTrainable)
-        assert issubclass(imagenet_trainable, ImagenetTrainable)
-        kwargs = dict(zip(kwargs_names, [imagenet_trainable, *tune.run.__defaults__]))
+        supervised_trainable = config.get("supervised_trainable", SupervisedTrainable)
+        assert issubclass(supervised_trainable, SupervisedTrainable)
+        kwargs = dict(zip(kwargs_names, [supervised_trainable, *tune.run.__defaults__]))
 
     # Check if restoring experiment from last known checkpoint
     if config.pop("restore", False):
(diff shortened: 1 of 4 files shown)


nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -293,11 +289,39 @@ def _stop(self):
...
 
-class SigOptImagenetTrainable(ImagenetTrainable):
+class SupervisedTrainable(BaseTrainable):
+    """
+    Trainable class used to train supervised machine learning experiments
+    with ray.
+    """
+
+    def _run_iteration(self):
+        """Run one epoch of training on each process."""
```
lucasosouza, line 308: Is the word epoch here still applicable?

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -379,9 +408,9 @@ def run(config):
         kwargs = dict(zip(kwargs_names, [SigOptImagenetTrainable,
                                          *tune.run.__defaults__]))
     else:
-        imagenet_trainable = config.get("imagenet_trainable", ImagenetTrainable)
-        assert issubclass(imagenet_trainable, ImagenetTrainable)
-        kwargs = dict(zip(kwargs_names, [imagenet_trainable, *tune.run.__defaults__]))
+        supervised_trainable = config.get("supervised_trainable", SupervisedTrainable)
```
lucasosouza, line 411: Why is the variable called supervised_trainable? Can we instead just call it trainable and then point to whatever trainable class the user wants?

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: It is because it's in the SupervisedTrainable.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: Good catch. Just updated.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -293,11 +289,39 @@ def _stop(self):
...
+            return ray.get(status)
+
+    def _process_result(self, result, pre_experiment_result=None):
+
+        # Aggregate initial validation results (before any training).
+        if pre_experiment_result is not None:
+            result["extra_val_results"].insert(0, (0, pre_experiment_result))
+
+
+class SigOptImagenetTrainable(SupervisedTrainable):
```
lscheinkman, line 324: Maybe `BaseTrainable` and `SupervisedTrainable ` should be on a different file from `SigOptImagenetTrainable` or `SigOptImagenetTrainable` could be renamed to `SigOptSupervisedTrainable `

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
lucasosouza: It is a matter of opinion, but I would still argue that it should be just `trainable` and not `ray_trainable`, since we discussed about removing explicit references to Ray whenever possible. 

We should be allowed to use other Trainables (considering the trainable is just a boilerplate function that call iterations from any class until a stopping criteria is met, and save the outcome at each step).

mvacaporale: I believe that strategy was discussed particularly for the Experiment classes. Here, the trainable will be passed to `ray.tune.run` so I don't think the same philosophy can be applied.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: The sipopt trainable uses `SigOptImagenetExperiment` which is quite tied to imagenet experiments (in terms of parameters it assumes we're using). I figured both of these could be made more general, but I've left it for later for now. After Lucas and I do further refactoring, we'll know more which is the best route.

PR #334, opened 2020-08-15
Revert "Initial Refactor for Experiment Trainables"

lucasosouza:
Reverts numenta/nupic.research#333

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
@@ -459,12 +459,6 @@ class ImagenetExperiment:
         else:
             return True
 
-    def pre_experiment(self):
-        """Run validation before training."""
-        if -1 in self.epochs_to_validate:
-            self.logger.debug("Validating before any training:")
-            return self.validate()
-
     def validate(self, loader=None):
         if loader is None:
             loader = self.val_loader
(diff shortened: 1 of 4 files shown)


lscheinkman:
@lucasosouza Please do not create branches on https://github.com/numenta/nupic.research
You should create a branch on your own fork

lucasosouza:
@mvacaporale 
This particular PR was created automatically by github when using reverse. We found a subtle bug on this while we were running some tests on Saturday, so I reversed.

We did create a temporary branch (RES-1630) to facilitate pair programming in remote setting, since we were both pushing commits to one branch. But it will be deleted very soon, after we are done with it Monday. Let me know if you still think it is an issue in this scenario.

lscheinkman:
For this particular PR you could use github to revert the changes on **your fork** (`lucassouza/nupic.research`) instead of **upstream** (`numenta/nupic.research`)  and later create a PR from **your fork** to **upstream** like any other PR.

As for collaboration, you could create a branch on your own **fork** (`lucassouza/nupic.research`) and he could just add your **fork** as a remote.

The reason for that is that everybody uses `numenta/nupic.research` as the **upstream** repository and when we download changes from the **upstream** repository we would also be downloading the changes to **your branches** which is not desirable.

Lets try to keep the **upstream** (`numenta/nupic.research`) repository clean from local changes

PR #335, opened 2020-08-15
Single instance fixes

lucasosouza:
Small fixes to single instance runner. When this was created, it solved two problems that at the time weren't functional in the main runner:
- running ray in single instance
- running several trials at once, to be able to do grid/random search through hyperparameter space.

The latter is fixed now, but I keep using single_instance for the first reason. This refactor is mainly to only use the Trial_Collection class only if the user specifically asks, maintaining compatibility with some of the early experiments I conducted on Imagenet. Not the prettiest code, but works alright.


lscheinkman:
@lucasosouza Do you need to run **ray** on a single instance?
How about "[/projects/imagenet/local_run.py](https://github.com/numenta/nupic.research/blob/master//projects/imagenet/local_run.py)" as an alternative?

lucasosouza:
Yes, I do it all the time, to test and debug new code, run small datasets, do some fast finetuning. I've never seen `local_run` before, that was probably created after I coded the single instance runner - so that code works with Ray and multiple GPUs as well?

lscheinkman:
> I've never seen `local_run` before, that was probably created after I coded the single instance runner - so that code works with Ray and multiple GPUs as well?

`local_run` was created a few months ago and it works just like ray on multiple GPU but only on a single node. 
Ray tune trials are converted to plain `dicts` (see [create_trials](https://github.com/numenta/nupic.research/blob/master//projects/imagenet/local_run.py#L56)) and ray remote calls are replaced with `torch.multiprocessing` (see [imagenet_run](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/imagenet/imagenet_run.py)). 
The logging and checkpointing is the same as ray (see [run_trial](https://github.com/numenta/nupic.research/blob/master//projects/imagenet/local_run.py#L119))

PR #336, opened 2020-08-17
Initial Refactor for Experiment Trainables (with fix)

mvacaporale:
This include all the changes from pull #333, with an additional minor fix
* Handle the case when `pre_experiment` returns a None result.

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
+++ b/nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
@@ -661,7 +661,8 @@ class ImagenetExperiment:
 
     @classmethod
     def aggregate_pre_experiment_results(cls, results):
-        return cls.aggregate_validation_results(results)
+        if all(results):
+            return cls.aggregate_validation_results(results)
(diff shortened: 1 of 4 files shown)


nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -653,6 +659,11 @@ def aggregate_validation_results(cls, results):
         result.update(aggregate_eval_results(results))
         return result
 
+    @classmethod
+    def aggregate_pre_experiment_results(cls, results):
+        if all(results):
```
lscheinkman, line 664: Is this something that only happens on `pre_experiment`?
Maybe this check should be placed in `aggregate_validation_results` instead to be safe.

mvacaporale: That's a fair question. The pre_experiment is optionally run (e.g. `-1 in epochs_to_validate`) while validate should return a dict, or at least that's the convention in the code right now.

PR #338, opened 2020-08-18
Refactoring for new SupervisedExperiment

mvacaporale:
The majority of this refactor is for creating datasets in the experiment classes. This is done through
* New class method `create_loaders`
* New class `ImagenetDataset` that has methods `get_train_dataset`, `get_val_dataset`, and `get_test_dataset`.
* New TorchvisionDataset class with the same methods


mvacaporale:
@lucasosouza Just a heads up, here is the PR we worked on

lucasosouza:
I will just add a comment it is a work in progress. We are breaking down into smaller PRs, and making sure at each PR all the unit and regression tests are cleared. So this is a rough first draft. Some things here have already been changed at commits that have yet not been pushed.

We will follow with more specific unit and regression tests as well.

mvacaporale:
I made the changes to the dataset manager classes (e.g. instead of ImagenetDataset it's now ImagenetDatasetManager). This PR is ready for review.

lucasosouza:
> I made the changes to the dataset manager classes (e.g. instead of ImagenetDataset it's now ImagenetDatasetManager). This PR is ready for review.

To add to the mix, how about DataManager to make it a little bit shorter?

mvacaporale:
> > I made the changes to the dataset manager classes (e.g. instead of ImagenetDataset it's now ImagenetDatasetManager). This PR is ready for review.
> 
> To add to the mix, how about DataManager to make it a little bit shorter?

I thought about that, but it particularly manages the datasets.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -66,14 +65,17 @@
...
+    "ImagenetExperiment",
+]
 
 
 # Improves performance when using fixed size images (224) and CNN
 cudnn.benchmark = True
 
 
-class ImagenetExperiment:
+class SupervisedExperiment:
```
lscheinkman, line 78: Is `SupervisedExperiment` specific for `resnet50 v1.5` on `Imagenet` or generic for any model on any dataset?
If it is generic then it should not be in the `imagenet_experiment.py` file

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -422,16 +438,11 @@ def create_train_dataloader(cls, config):
         )
 
     @classmethod
-    def create_validation_dataloader(cls, config):
+    def create_validation_dataloader(cls, dataset, config):
```
mrcslws, line 441: This is making it more difficult to write a test scripts that run validation. Previously we could say `loader = experiment_class.create_validation_loader(config)`. Now we have to 
figure out how to instantiate our own dataset.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -392,20 +395,33 @@ def create_lr_scheduler(cls, config, optimizer, total_batches):
...
+
+        dataset_manager_class = config.get("dataset_manager_class", None)
+        if dataset_manager_class is None:
+            raise ValueError("Must specify 'dataset_manager_class' in config.")
+        dataset_args = config.get("dataset_args", {})
+        dataset_manager = dataset_manager_class(**dataset_args)
+
+        train_set = dataset_manager.get_train_dataset()
+        val_set = dataset_manager.get_val_dataset()
+        test_set = dataset_manager.get_test_dataset()
```
lscheinkman, line 409: Maybe you could refactor this code as a class method as well (`create_dataset_manager`). It would allow other code to create `dataset_manager` accessing the experiment's datasets without instantiating dataloaders loaders. This could be useful for testing and other reporting scripts.

nupic/research/frameworks/pytorch/datasets/common.py
```diff
@@ -0,0 +1,80 @@
...
+            train=False,
+            transform=transform,
+            download=False,
+        )
+
+    def get_train_dataset(self):
+        return self.train_dataset
+
+    def get_val_dataset(self):
+        return self.val_dataset
```
mrcslws, line 67: Could `SupervisedExperiment.create_validation_dataloader` call `DatasetManager.create_validation_dataset` directly? (To solve the "test script" problem I mention in a different comment)

In fact, is there a reason to ever need to instantiate a DatasetManager? It's a bit of a red flag that we're instantiating managers then never storing them. Maybe these classmethods should just be functions that return a dataset -- is there a reason we need a dedicated "Manager" class? This seems like a job for factory-style functions, similar to how the `common_layers.py` contains factory functions for layers. (Conveniently, each of those hypothetical functions could simply be called a "dataset" since they are dataset constructors.)

nupic/research/frameworks/pytorch/models/common.py
```diff
@@ -0,0 +1,41 @@
...
+#  You should have received a copy of the GNU Affero Public License
+#  along with this program.  If not, see http://www.gnu.org/licenses.
+#
+#  http://numenta.org/licenses/
+#
+
+import numpy as np
+import torch.nn as nn
+
+from nupic.torch.modules import Flatten
```
lscheinkman, line 24: pytorch 1.6 introduced [torch.nn.Flatten](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/flatten.py). We should start using pytorch version instead of `nupic.torch`

nupic/research/frameworks/pytorch/datasets/common.py
```diff
@@ -0,0 +1,80 @@
...
+            raise ValueError(f"{dataset_name} not available.")
+        transform = self.base_transform(*DATASETS_STATS[dataset_name])
+
+        # TODO: rename data to dataset_dir
+        dataset_class = getattr(datasets, dataset_name)
+        self.train_dataset = dataset_class(
+            root=os.path.expanduser(data_dir),
+            train=True,
+            transform=transform,
+            download=False,
```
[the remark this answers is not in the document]
mvacaporale: I made this change. We were hesitant because the user has to be careful when downloading in the multi-worker setting - it will throw an error when they try to all write to the same directory. I added a warning in the docstring for this behavior.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: It is generic. I've yet to do the renaming of all the files, however (including the imagenet directory to yael of whatever else we decide). If it's alright, the renaming will happen soon in a future PR.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: Per your comment and Marcus's, there's now a create_datasets_func that you pass to the config.

nupic/research/frameworks/pytorch/models/common.py
mvacaporale: Good point, changed.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: This is not neccessarily true. There's a new class method called as `experment_class.create_loaders` that will return train_loader and val_loader (and test_loader if any).

nupic/research/frameworks/pytorch/datasets/common.py
mvacaporale: This is good feedback. I guess there isn't a need for a whole datamanager class. I've refactored things to call a `create_datasets_func`. I returns the train and val datasets (and a test set if any).

nupic/research/frameworks/pytorch/datasets/common.py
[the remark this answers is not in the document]
mvacaporale: Is there a way to better manage this with file locking?

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: If you feel strongly about this, I can also do all the renaming now, ofc.

nupic/research/frameworks/pytorch/datasets/common.py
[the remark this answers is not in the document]
lscheinkman: Yes, here is an example from ray
https://docs.ray.io/en/latest/using-ray-with-pytorch.html

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lscheinkman: It is ok to rename the file in the future but you should at least update the docstring

lscheinkman:
@mvacaporale Make sure to run the regression tests before merging

lucasosouza:
> For our current scenarios, I still think it would still be possible to mimic torchvision's dataset functions more. Rather than having a "create_datasets" function, you could have factory functions or classes similar to `torchvision.datasets.CIFAR10`, `torchvision.datasets.MNIST`, etc., and the SupervisedExperiment could call the function (or constructor) twice with `train=True` and `train=False`. Essentially we could design the API so that we can pass torchvision datasets directly into the experiment config (though we'd often still use the wrapper).
> 
> You guys have spent more time thinking about what is needed for continual learning experiments, so I defer to you.

That is a good point, and that was the initial idea in the initial design proposal. Maybe for a next PR we can think about still having a wrapper, but subclassing from pytorch Dataset class instead. 

The wrapper is still required because there are specific things to each dataset that we want to define. For example, even for simple torchvision datasets we want to define some base transforms for each dataset, so it is standardized in the codebase. The experiment should not contain any dataset logic in order to be compatible with multiple datasets. 

We also didn't think it was a good idea to ask the user to define these transforms every time in the config file.

mrcslws:
Totally agree, the wrapper is useful. I'm just suggesting that we're being a good PyTorch citizen (and making our code easier to understand and compatible with other people's datasets) if we choose a design that supports using the torchvision datasets directly. (This doesn't mean we 100% need to do this, but it is desirable.)

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: Can do.

nupic/research/frameworks/pytorch/models/common.py
```diff
@@ -0,0 +1,39 @@
...
+#  http://numenta.org/licenses/
+#
+
+import numpy as np
+import torch.nn as nn
+
+
+class StandardMLP(nn.Module):
+
+    def __init__(self, input_size, num_classes):
```
subutai, line 27: Shouldn't we provide number of hidden units for each layer as a parameter here?

lucasosouza:
tagging @karangrewal so he can be aware of the changes as well, since it affects RES-1631 (meta continuous learning)

nupic/research/frameworks/pytorch/models/common.py
lucasosouza: Yes, you are right. This was the simplest network possible to run some tests. But it has been fixed already in future commits, this will look better in the next PR.

mvacaporale:
Note: The regression tests pass.

PR #339, opened 2020-08-21
Continual Learning Experiment (RES-1630-b)

lucasosouza:
@mvacaporale @karangrewal 

Implements base class for ContinualLearning. Will be followed by a specific mixin to implement a standard method (EWC), to ensure it can replicate the results from the original codebase, and tests for ContinualLearningExperiment.

**May change significantly as more ContinualLearning algorithms are implemented.**

Some of the main changes:
- Refactor dataset loading to make it compatible with pytorch Dataset signature, following @mrcslws suggestion
- ContinualLearningExperiment, with an outer loop over tasks, no validation between epochs and validation between tasks.
- A sampler that can handle changing between tasks. Distributed version available only for training now
- Evaluation metrics specific to ContinualLearning: we debated a few ways of doing this, and settled on using a "mandatory mixin" to clean the logic from ContinualLearningExperiment - any suggestions on a different approach are welcome. Please see the config file in nupic.hardware for more details on how it is defined.


nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+
+        self.num_classes_per_task = math.floor(self.num_classes / self.num_tasks)
+
+        # applying target transform depending on type of CL task
+        # task - we know the task, so the network is multihead
+        # class - we don't know the task
+        self.cl_experiment_type = config.get("cl_experiment_type", "class")
+        if self.cl_experiment_type == "task":
+            # override the target transform
+            self.dataset_args["target_transform"] = (
```
mvacaporale, line 955: Would this need to be passed to the validation dataset as well? I don't think this happens as of now.

Also, there's a lurking error where the code now assumes "target_transform" is accepted to your dataset constructor. A more general way to do this may be 
1. create your datasets
2. assert they are a VisionDataset if `cl_experiment_type == "task"`
3. and then modify their `target_transform`
That's a bit sly though. 

You could also possibly throw a warning when you modify the `dataset_args`, just to make sure the user knows the argument is being overridden and that it might not work with the `dataset_class` they've given.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+        """
+        Whether or not the experiment should stop. Usually determined by the
+        number of epochs but customizable to any other stopping criteria
+        """
+        return self.current_task >= self.num_tasks
+
+    def run_task(self):
+        """Run outer loop over tasks"""
+        # configure the sampler to load only samples from current task
+        print("Training...")
```
mvacaporale, line 977: Use the self.logger over a print statement

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+        # run epochs, inner loop
+        # TODO: return the results from run_epoch
+        for _ in range(self.epochs):
+            self.run_epoch()
+
+        ret = {}
+        for metric in self.evaluation_metrics:
+            eval_function = getattr(self, metric)
+            temp_ret = eval_function()
+            print(temp_ret)
```
mvacaporale, line 989: Another print statement

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+            for k, v in temp_ret.items():
+                ret[f"{metric}__{k}"] = v
+
+        # TODO: fix this
```
mvacaporale, line 993: Maybe expand on the todo statement. It's not clear what should be fixed

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+            extra_val_results=[],
+        )
+
+        self.current_task += 1
+        return ret
+
+    @classmethod
+    def aggregate_results(cls, results):
+        """Run validation in single GPU"""
+        print("Aggregated results")
```
mvacaporale, line 1007: You should check for all your print statements

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -393,6 +410,31 @@ def _process_result(self, result, pre_experiment_result=None):
                     pprint(result)
 
 
+class DebugTrainable(Trainable):
```
mvacaporale, line 413: You should remove all debugging code

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -442,20 +484,28 @@ def run_single_instance(config):
...
-    kwargs = dict(zip(kwargs_names, [SupervisedTrainable, *tune.run.__defaults__]))
+    kwargs = dict(zip(kwargs_names, [ray_trainable, *tune.run.__defaults__]))
     # Update`tune.run` kwargs with config
     kwargs.update(config)
     kwargs["config"] = config
     # Update tune stop criteria with config epochs
     stop = kwargs.get("stop", {}) or dict()
     epochs = config.get("epochs", 1)
-    stop.update(training_iteration=epochs)
+    num_tasks = config.get("num_tasks", None)
```
mvacaporale, line 500: This was also debugging code I believe

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
lucasosouza: I am keeping this class for future debugging, it is rather useful

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lucasosouza: Should I just change all print statements to logger.debug instead, what do you think? This is still in development, I don't want to be removing print statements and putting them back everytime we push a new PR on this (actually I've already removed 90% of the print statements we had).

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
lucasosouza: Not really, this is required to run ContinualLearningExperiment in single instances. Arguably there are better ways of doing this, but it is fine as it is now.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lucasosouza: `dataset_args` applies equally to both training and validation datasets (which is probably an issue, we should find a way of setting different transforms and target_transforms to both). 

Good catch on the error, I will work on the remaining changes. Truth is we never really tested the `task` approach until now, only the `class` - these will come up more obviously once we implement EWC and start comparing results.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: Part of the process could be simply not committing your print statement as well. I think we should try and keep master as tidy as possible.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: That's right, sorry. This works then

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
mvacaporale: That's fair. I don't have a better name. It for debugging after all. If anything, maybe we should move it to another file in the future. Maybe add a TODO for this.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -395,21 +408,30 @@ def create_lr_scheduler(cls, config, optimizer, total_batches):
...
+    @classmethod
+    def create_train_sampler(cls, dataset, config):
+        if config.get("distributed", False):
+            sampler = DistributedSampler(dataset)
+        else:
+            sampler = None
+        return sampler
 
     @classmethod
     def create_train_dataloader(cls, dataset, config):
```
mrcslws, line 437: Now that you've changed this to use simpler dataset factory functions, I think it's worth changing these methods back to the previous signatures, `create_train_dataloader(cls, config)` and `create_validation_dataloader(cls, config)` (without the `dataset` input) so that they're easier to use in test  scripts, quantization scripts, etc. We don't need the new `create_loaders` method any more. These methods can create their own datasets easily.

This is only tangentially related to this PR, so feel free to do it in a later PR. (If you agree.)

lucasosouza: I don't have a strong opinion about this, I would be fine either way. Maybe @mvacaporale can chip in. You are right that as we refactored the need for a `create_loaders` function may no longer be there, and I see your point on having a function that doesn't take dataset as argument as being easier to use. I honestly don't remember why we decided to go with that approach it in the first place.

mvacaporale: I think the motivation was for the `create_loaders`. We had one dataset function that made both the train and val sets, and so we needed one function to create the loaders. But this is not needed anymore as you both pointed out.

I can make the change after this PR is merged.

mvacaporale:
Approved, baring the aforementioned discussed changes.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
 
     @classmethod
     def get_execution_order(cls):
+        exp = "SupervisedExperiment"
```
lscheinkman, line 897: Why are you introducing this constant and the extra string concatenations?

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
```diff
@@ -869,47 +894,173 @@ def get_free_port(self):
...
+
+        # defines how many classes should exist per task
+        num_tasks = config.get("num_tasks", 1)
+        num_classes = config.get("num_classes", None)
+        assert num_classes is not None, "num_classes should be defined"
+        num_classes_per_task = math.floor(num_classes / num_tasks)
+
+        task_indices = defaultdict(list)
+        for i in range(num_tasks):   # 1 to 5
+            for j in range(num_classes_per_task):  # 1 to 200
```
lscheinkman, line 1033: What do you mean by comments next to `num_tasks` to be `1 to 5` and `num_classes_per_task` to be `1 to 200`?

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
```diff
@@ -442,20 +484,28 @@ def run_single_instance(config):
     config["reuse_actors"] = False
     config["dist_port"] = get_free_port()
 
+    # set trainable
+    ray_trainable = config.get("ray_trainable", SupervisedTrainable)
+    # assert issubclass(ray_trainable, BaseTrainable)
```
lscheinkman, line 489: Why did you remove this assert?

nupic/research/frameworks/pytorch/imagenet/mixins/reduce_lr_after_task.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+
+class ReduceLRAfterTask:
+    """
+    Freeze after params after task k.
+    """
+    def setup_experiment(self, config):
```
lscheinkman, line 27: Could you add a docstring describing the config parameters?

nupic/research/frameworks/pytorch/models/common_models.py
```diff
@@ -0,0 +1,88 @@
...
+            ])
+        layers.append(nn.Linear(hidden_sizes[-1], num_classes))
+
+        self.classifier = nn.Sequential(*layers)
+
+    def forward(self, x):
+        return self.classifier(x)
+
+
+class SmallCNN(nn.Module):
```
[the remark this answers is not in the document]
lucasosouza: Good point. It is very specific, I will change it to OmniglotCNN.

nupic/research/frameworks/pytorch/dataset_utils/samplers.py
```diff
@@ -0,0 +1,101 @@
...
+        self.task_indices = task_indices
+        self.set_active_tasks(0)
+
+    def set_active_tasks(self, tasks):
+        """Accepts index for task or list of indices"""
+        self.active_tasks = tasks
+        if not isinstance(self.active_tasks, Iterable):
+            self.active_tasks = [tasks]
+        self.indices = np.concatenate([self.task_indices[t] for t in self.active_tasks])
+        self.num_samples = math.ceil(len(self.indices) * 1.0 / self.num_replicas)
```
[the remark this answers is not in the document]
lucasosouza: Some of these lines were copied from pytorch samplers. I guess the multiplication there was to convert from int to float, which is not really required in python 3. I will remove it.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lucasosouza: This was an idea by @mvacaporale to make the code look cleaner, there is no reason to repeat "ImagenetExperiment" dozens of times. It also helps avoid errors from typos. I see no harm in this.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lucasosouza: That shouldn't be there, I will remove it.

nupic/research/frameworks/pytorch/imagenet/imagenet_tune.py
lucasosouza: I will reintroduce as `assert issubclass(ray_trainable, Trainable)`. We now can have trainables that are not subclassed from `BaseTrainable` but from `Trainable` directly.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
lscheinkman: It does not look cleaner to me and there is a performance cost of 25+ extra string concatenation

lucasosouza:
Thanks for all the feedback, that is great. I will push a new commit with all the requested changes tomorrow.

nupic/research/frameworks/pytorch/imagenet/imagenet_experiment.py
mvacaporale: It's cleaner in terms of non-wrapping lines. And I think the performance cost is small, right?

I'm fine either way though.

lscheinkman: If wrapping lines is the issue, I would `# noqa: E501` to the method before adding extra string concatenations

lucasosouza: Style can be argued. But the efficiency argument doesn't hold. This is O(1) and only executed once at initialization, we are talking about nanoseconds here. 

Given the differences in opinion, I will abide by the opinion of whoever coded in the first place. @mrcslws git blame says it is your code. If you are not against the changes, I will keep them.

PR #340, opened 2020-08-21
Refactor `create_model` to be more general

mvacaporale:
As we refactor the experiment class, it’s important to have a general `create_model` function. Previously, it called `init_resnet50_batch_norm` which reinitialized the resent batchnorm layers and any linear layers as well. This PR generalizes this functionality by
* Passing an `init_model_fn` to `create_model`; applied via model.apply
* Refactoring `init_resnet50_batch_norm` to work via model.apply
* Modify `create_model` classmethod of `ImagementExperiment` to set `init_model_fn=init_resnet50_batch_norm` when `init_batch_norm=True`


mvacaporale:
Closing this PR for later work. I'll most likely go the mixin route with a future PR.

PR #342, opened 2020-09-01
Storing the main parser arguments in one location

mvacaporale:
We reuse many of the same arguments for parsing commands for our experiment. This is done in our `run*.py` files in `nupic.research.projects` and other places as well.

This refactor introduces `parser_utils.py` to store an importable `MAIN_PARSER` and `RAY_PARSER` to be used as parent parsers when needed.

As well, in the same file, there's logic to process the args and modify the config as appropriate - e.g.`insert_experiment_mixin`

### DIFF
--- a/nupic/research/frameworks/pytorch/imagenet/parser_utils.py
+++ b/nupic/research/frameworks/pytorch/imagenet/parser_utils.py
@@ -33,8 +33,11 @@ from nupic.research.frameworks.sigopt.sigopt_experiment import SigOptImagenetExp
 from nupic.research.frameworks.wandb import ray_wandb
 
 __all__ = [
+    "DEFAULT_PARSERS",
     "MAIN_PARSER",
     "RAY_PARSER",
+    "SIGOPT_PARSER",
+    "WANDB_PARSER",
     "process_args",
     "insert_experiment_mixin",
 ]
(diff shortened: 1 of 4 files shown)


mvacaporale:
@lucasosouza Just a heads up, this is a PR we've talked about.

Question for you: I set `load_code_from_local=False` in `run_trial_single_instance`. This helped prevent an error when ray said it couldn't retrieve the `experiment_class` from the local module after it was modified by `insert_experiment_mixin`. Everything seemed to run just fine given this change. Do you have any thoughts if it will negatively affect anything?

nupic/research/frameworks/pytorch/imagenet/parser_utils.py
```diff
@@ -0,0 +1,165 @@
...
+import torch
+from wandb import util
+
+from nupic.research.frameworks.dynamic_sparse.common.ray_custom_loggers_2 import (
+    DEFAULT_LOGGERS,
+)
+from nupic.research.frameworks.pytorch.imagenet import mixins
+from nupic.research.frameworks.sigopt.sigopt_experiment import SigOptImagenetExperiment
+from nupic.research.frameworks.wandb import ray_wandb
+
```
lscheinkman, line 34: Could you add `__all__` list here exporting the parsers and the helper functions?

nupic/research/frameworks/pytorch/imagenet/parser_utils.py
```diff
@@ -0,0 +1,165 @@
...
+MAIN_PARSER.add_argument("--profile", action="store_true",
+                         help="Enable cProfile tracing")
+MAIN_PARSER.add_argument("--profile-autograd", action="store_true",
+                         help="Enable torch.autograd.profiler.profile during training")
+MAIN_PARSER.add_argument("--wandb", action="store_true",
+                         help="Enable logging through wandb.")
+MAIN_PARSER.add_argument("--wandb_resume", action="store_true",
+                         help="Resume logging through wandb.")
+MAIN_PARSER.add_argument("-t", "--create_sigopt", action="store_true",
+                         help="Create a new sigopt experiment using the config")
```
lscheinkman, line 76: Maybe you could separate `wandb`, `sigopt` and any other 3rd party library options into different parsers, just like what you did for `RAY_PARSER` and keep the `MAIN_PARSE` clean from 3rd party dependencies

mvacaporale: Done. I also introduced a `DEFAULT_PARSERS` list with all of them. Let me know if you think this is reasonable (it looks cleaner to me to have to import only one thing as opposed to four individual parsers when possible).

PR #344, opened 2020-09-09
Rename Experiments Platform to Vernon

mvacaporale:
This PR reorganizes and relabels our generalized set of tools for running and analyzing experiments. What use to be `nupic.research.frameworks.pytorch.imagenet` is now `nupic.research.frameworks.vernon`. 
* Imports are fixed as appropriate.
* Some subfiles were renamed (e.g. `experiment_search.py` to `search.py`).
* To run configs, one may now use either `veron.run` or `vernon.run_with_raytune` - the latter accepts a new boolean argument `use_single_instance`.


nupic/research/frameworks/vernon/parser_utils.py
```diff
@@ -89,7 +89,7 @@
 RAY_PARSER.add_argument("-s", "--with-server", action="store_true",
                         help="Start Ray Tune API server")
 RAY_PARSER.add_argument("--single_instance", action="store_true",
-                        help="Uses single instance run method")
+                        help="Uses single instance run method", default=False)
```
lscheinkman, line 92: You dont need to add `default=False` when using `action="store_true"`.
See https://docs.python.org/3.7/library/argparse.html?highlight=store_true#action

projects/gsc/run.py
```diff
@@ -50,7 +47,5 @@
 
     if config is None:
         pass
-    elif "single_instance" in args:
-        imagenet_tune.run_single_instance(config)
     else:
-        imagenet_tune.run(config)
+        vernon.run_with_raytune(config, use_single_instance=args.single_instance)
```
lscheinkman, line 51: `single_instance` is created by `RAY_PARSER` but it is handled outside the `vernon.run_with_raytune`. 
This may lead to inconsistencies when the caller uses `RAY_PARSER` but does call `vernon.run_with_raytune` with the extra parameter.

projects/imagenet/run.py
```diff
@@ -49,4 +46,4 @@
     config = process_args(args, config)
 
     if config is not None:
-        imagenet_tune.run(config)
+        vernon.run_with_raytune(config)
```
lscheinkman, line 49: Missing handling of `single_instance` argument created by `RAY_PARSER`. 
See my previous comment

lscheinkman (commented):
Could you also run the regression tests?

projects/gsc/run.py
mvacaporale: Could you clarify this a bit more? I'm not understanding the implication.

mvacaporale:
> Could you also run the regression tests?

Will do.

lucasosouza (approved):

projects/gsc/run.py
mvacaporale: I think I get it. You saying that instead of passing `use_single_instance=args.single_instance` to `vernon.run_with_raytune`, I should handle the argument inside `vernon.run_with_raytune` via the config as in 
```
def run(config):

    if config.get("use_single_instance", False):
        return run_single_instance(config)
...
```

Is this correct?

lscheinkman: yes

mvacaporale:
Regression tests pass. I will merge this PR now.

PR #353, opened 2020-09-25
Enable using latest wandb (Remove dependency on WandbLogger)

mrcslws:
Trivial solution: inline the WandbLogger. I think it's better to implement our own logger anyway, rather than subclassing, so that we get direct control over how we use wandb's API.

Maybe in the future we'll want to fork/subclass `ray.tune.integration.wandb`, but in the meantime this change is an improvement.

### DIFF
--- a/nupic/research/frameworks/wandb/ray_wandb.py
+++ b/nupic/research/frameworks/wandb/ray_wandb.py
@@ -28,6 +28,7 @@ from copy import deepcopy
 from pprint import pformat
 
 import wandb
+from ray import tune
 from ray.tune.utils import flatten_dict
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


mvacaporale (approved):
Looks good. Please test it, of course, to make sure logging works as you'd expect (we don't have any unit tests for wandb yet to prevent regressions).

mvacaporale:
Curious, what features were you hoping to use in the latest version of wandb?

mvacaporale:
Later on, I'd like to update ray to 0.8.7 so we can use the `ray.tune.integration.wandb`. There's also `ray.tune.integration.wandb.wandb_mixin` to help initialize wandb outside of a trainable.

mrcslws:
Yes, I tested it. There wasn't a specific feature I wanted to use, it just seemed worthwhile to keep up-to-date, getting any bugfixes, perf improvements, etc. Also, I think the code is better this way regardless. (In this case, subclassing was making the code harder to read, not easier)

PR #355, opened 2020-09-28
Remove dependency on ray in vernon

mvacaporale:
In `nupic.research.frameworks.vernon.__init__.py` there were imports that required ray. This caused issues when pickle5 was imported prior to this module. 

This PR removes this imports and moves the run and trainable files to maintain an easy to use API.


mvacaporale:
Running regression tests now.

projects/imagenet/local_run.py
```diff
@@ -35,9 +35,9 @@
 import torch.multiprocessing as multiprocessing
 
 from experiments import CONFIGS
-from nupic.research.frameworks import vernon
 from nupic.research.frameworks.vernon import ImagenetExperiment
 from nupic.research.frameworks.vernon.parser_utils import MAIN_PARSER, process_args
+from nupic.research.frameworks.vernon.run import run, terminate_processes
```
lucasosouza, line 40: Maybe we can look for a better name later so we don't have a `from run import run` situation. Although it is quite common in python like `time.time` or `copy.copy`.

lucasosouza (approved):
Looks ok to me. Was this the issue blocking the regression tests?

projects/imagenet/local_run.py
mvacaporale: Good point. That's why I had it the other way before. There's a trick to make `run` a callable module, but I figured this would be simpler for now.

lscheinkman: @mvacaporale You should remove the `import ray` and the `FIXME` from this file and try to run any experiment.
https://github.com/numenta/nupic.research/pull/355/files#diff-87f6f72f74eee2681c045d1824029eafR33-R34

lscheinkman:
@mvacaporale  It still fails with `ImportError: Ray must be imported before pickle5 because Ray requires a specific version of pickle5 (which is packaged along with Ray).` See stack trace below:
```
from nupic.research.frameworks.vernon.parser_utils import MAIN_PARSER, process_args
  File "~/nta/nupic.research/nupic/research/frameworks/vernon/parser_utils.py", line 28, in <module>
    exitcode = _main(fd)
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/multiprocessing/spawn.py", line 114, in _main
    from nupic.research.frameworks.dynamic_sparse.common.ray_custom_loggers_2 import (
  File "~/nta/nupic.research/nupic/research/frameworks/dynamic_sparse/common/ray_custom_loggers_2.py", line 33, in <module>
    prepare(preparation_data)
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/multiprocessing/spawn.py", line 225, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/multiprocessing/spawn.py", line 277, in _fixup_main_from_path
    run_name="__mp_main__")
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "~/nta/nupic.hardware/projects/imagenet/local_run.py", line 28, in <module>
    from nupic.research.frameworks.vernon.parser_utils import MAIN_PARSER, process_args
  File "~/nta/nupic.research/nupic/research/frameworks/vernon/parser_utils.py", line 28, in <module>
    from nupic.research.frameworks.dynamic_sparse.common.ray_custom_loggers_2 import (
  File "~/nta/nupic.research/nupic/research/frameworks/dynamic_sparse/common/ray_custom_loggers_2.py", line 33, in <module>
    from ray.tune.logger import CSVLogger, JsonLogger, Logger
    from ray.tune.logger import CSVLogger, JsonLogger, Logger
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/site-packages/ray/__init__.py", line 12, in <module>
  File "~/miniconda3/envs/nupic.hardware/lib/python3.7/site-packages/ray/__init__.py", line 12, in <module>
    raise ImportError("Ray must be imported before pickle5 because Ray "
ImportError: Ray must be imported before pickle5 because Ray requires a specific version of pickle5 (which is packaged along with Ray).
    raise ImportError("Ray must be imported before pickle5 because Ray "
ImportError: Ray must be imported before pickle5 because Ray requires a specific version of pickle5 (which is 
```

mvacaporale:
> ImportError: Ray must be imported before pickle5

I have an idea on how to adjust `parser_utils.py` so that it doesn't import ray, but some configs in `nupic.research/projects/imagenet` rely on `ray.tune.grid_search` and so they import ray. I'm not sure the best way to deal with this issue. Any thoughts?

lscheinkman:
`tune.grid.search` function call can be replaced by a `dict`. For example:
```
tune.grid.search(["relu", "tanh"])
```
Is equivalent to:
```
{"grid_search": ["relu", "tanh"]}
```

See https://docs.ray.io/en/latest/_modules/ray/tune/suggest/variant_generator.html#grid_search

mvacaporale:
I saw that too. But there could be other sources of ray imports that won't have a quick fix. For instance, someone may define their own custom trainable and pass that through the config. This may not be in `basy.py` but some other config file imported when calling `from experiments import CONFIGS`.

I see the value in adjusting the imports of Vernon to ensure anyone wanting to use it without ray can do so without an import error. However, the problem within the configs extends beyond Vernon. I'm not sure there's a reasonable fix for it. If anything, one should be aware of this issue when defining the configs. Potentially, there should be an `experiments_with_ray` folder that's kept separate.

lscheinkman:
Do you see anyone using anything other than `tune.grid.search` in the configs right now?

lscheinkman:
My only request is that we keep `vernon` independent of `ray`.  
If the `CONFIGS` themselves are dependent on `ray` then we should separate `CONFIGS` into multiple config files, one with `ray` and the other without `ray`. This way the `run.py` should decide which config to use, but `vernon` itself should be independent of `ray`

PR #358, opened 2020-09-30
Compute and backpropagate error loss and complexity loss separately

mrcslws:
Before this change, complexity loss functions were memory-constrained, since storing the forward pass computation graph uses nearly all available memory. This competition for memory is totally unnecessary; complexity loss can be computed and backpropagated after this computation graph has been freed.


mrcslws:
I ran and profiled an experiment that uses complexity loss. Everything looks good.

lscheinkman (approved):

PR #373, opened 2020-10-12
API changes: train_model_func, evaluate_model_func, epochs_to_validate

mrcslws:
This change deprecates `train_model_func` and `evaluate_model_func`, modifies the default for `epochs_to_validate`, and adds a `validate_immediately` bool that is equivalent to including -1 in `epochs_to_validate`.

See commit descriptions for more info.


mrcslws:
I'll resubmit a different version of these changes after the other refactoring.

PR #382, opened 2020-10-14
Test regular network on routing task

karangrewal:
This introduces `regular_network.py` which trains a single-layer network on the routing task (in which the input and context vectors are concatenated as input to the network).

**Results:** The single-layer network's loss stops decreasing quite soon. Furthermore, there's no significant difference in mean abs. error on the dimensions that get masked (i.e., output multiplied by 0) vs. those that don't. The mean abs. error never goes below 0.05, whereas it does in the case of the learned dendrite module. This suggests the single-layer network doesn't learn the routing task.

### DIFF
--- a/nupic/research/frameworks/dendrites/routing/routing.py
+++ b/nupic/research/frameworks/dendrites/routing/routing.py
@@ -19,11 +19,92 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+from random import randint
+
 import torch
+from torch.utils.data import Dataset
 
 from nupic.torch.modules import SparseWeights
 
 from .utils import generate_random_binary_vectors
+
+
+class RoutingDataset(Dataset):
+    """
+    A dataset class for generating input-target pairs for the routing test specifically
+    for a linear network with a single layer, where the inputs are random vectors
+    sampled from U[-2, 2) each paired a binary sparse context vector
+    """
+
+    def __init__(
+        self,
+        routing_function,
+        input_size,
+        context_vectors,
+        device,
+        concat=False,
+        dataset_size=1e4,
+        x_min=-2.0,
+        x_max=2.0,
+    ):
+        """
+        :param routing_function: the random routing function
+        :param input_size: the number of dimensions in the input to the routing
+                           function and test module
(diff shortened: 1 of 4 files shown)


nupic/research/frameworks/dendrites/routing/utils.py
```diff
@@ -103,3 +103,132 @@ def get_gating_context_weights(output_masks, context_vectors, num_dendrites):
...
+    with torch.no_grad():
+        for data_and_context, target in loader:
+
+            data_and_context = data_and_context.to(device)
+            target = target.to(device)
+
+            output = model(data_and_context)
+            loss += criterion(output, target, reduction="sum")
+
+            # Report mean absolute error for masked and unmasked indices separately
```
subutai, line 220: What are these lines doing, and why is it different for non-dendrite models? There's a lot of duplicated code. Ideally the training and evaluation loops should use the exact same functions for both dendrite and non-dendrite models. The only difference is whether the context vectors are concatenated or not.

nupic/research/frameworks/dendrites/routing/routing.py
```diff
@@ -19,13 +19,94 @@
...
+        # To retrieve an input-context-target pair, first generate noise from a uniform
+        # distribution as input (where the bounds of the distribution were specified in
+        # the `__init__` method), and take the routing function's output on said input
+        # using any of its output masks
+
+        if idx > self.size:
+            raise IndexError("Index {} is out of range".format(idx))
+        torch.manual_seed(idx)
+
+        x = self.alpha * torch.rand((self.input_size,)) - self.beta
```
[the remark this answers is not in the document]
karangrewal: Yes, I will fix this. As an aside, I don't think this will make a big difference in terms of results, since right now we're sampling inputs i.i.d. from U(2, 6) instead of U(-2, 2).

nupic/research/frameworks/dendrites/routing/utils.py
karangrewal: Ok, we don't need all those metrics, so I just included mean absolute error (in addition to `criterion` provided as an arg). Also, I combined the train and evaluate functions to remove duplicate code.

subutai (approved):

PR #386, opened 2020-10-16
Dendritic activations visualization tool

karangrewal:
The file `nupic/research/frameworks/dendrites/utils.py` introduces a function used for visualizing dendritic activations in a heatmap using wandb, and `nupic/research/frameworks/dendrites/visualize_demo.py` gives a quick demo on how to use this tool.


nupic/research/frameworks/dendrites/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        activations = torch.abs(activations)
+    activations = activations.detach().cpu().numpy()
+
+    num_contexts = context_vectors.size(0)
+    num_dendrites = dendritic_weights.size(0)
+
+    x_labels = ["context {}".format(j) for j in range(num_contexts)]
+    y_labels = ["dendrite {}".format(j) for j in range(num_dendrites)]
+
+    wandb.init(name=run_name, project=project_title)
```
mvacaporale, line 62: It would be best to make this function simpler in its purpose. In programming, it often helps to have objects (like functions, classes, etc.) serve one role, or at least reduce the number of roles they do serve.

In this case `visualize_dendritic_activations` could simply make and return a plot via an axes or figure object.

nupic/research/frameworks/dendrites/visualize_demo.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    dendritic_network = dendritic_network.to(device)
+    context_vectors = context_vectors.to(device)
+
+    dendritic_weight_tensor = dendritic_network.segments.weights.data
+
+    # Visualize the absolute activations from just the first output unit's dendritic
+    # weights using Weights & Biases
+    visualize_dendritic_activations(
```
mvacaporale, line 62: Something else to keep in mind: When using wandb, the typical practice is to init wandb in your desired process. And then call `wandb.log` on the fly.

In this case, it's simply calling wandb at the start of the demo.
```
def run_wandb_demo():
    wandb.init(...)
    ...
    plot = visualize_dendritic_activations(..).
    wandb.log({"my_plot", plot))
```

For a training run it could be something like this
```
def do_training(...):

    for e in range(epochs):
         # <training happens>
         ...
         
         if wandb.run is None:   # only log is wandb.init was called
             wandb.log(...)

if __name__ == "__main__"":
    # <setup experiment>
    wandb.init(...)
    ...
    do_training()
```

mvacaporale: One other thing to note: `wandb` needs a place to store results, as you know. A demo or test code should not, ideally, create unnecessary files and folders. You can avoid this by using tempfile:
```
import tempfile
tempdir = tempfile.TemporaryDirectory()

...

wandb.init(..., dir=tempdir.name)

...

tempdir.cleanup()
```

mvacaporale (changes requested):

nupic/research/frameworks/dendrites/utils.py
karangrewal: Fixed - the function now returns the heatmap object.

nupic/research/frameworks/dendrites/visualize_demo.py
karangrewal: I moved `wandb.init` to the script that gets called, but regarding your 2nd point on using `tempfile`, I don't think nupic.research supports this library (since it's not listed in requirements.txt). Thoughts on how to proceed?

subutai (approved):

nupic/research/frameworks/dendrites/visualize_demo.py
```diff
@@ -0,0 +1,79 @@
...
+    # unit's dendritic activations
+    visual = plot_dendritic_activations(
+        dendritic_weights=dendritic_weight_tensor[0, :, :],
+        context_vectors=context_vectors,
+        mask_values=mask_values,
+        use_absolute_activations=True
+    )
+
+    # Plot heatmap of dendrite activations using wandb
+    wandb.init(name="Output unit 0", project="Dendrites demo")
```
mvacaporale, line 74: Maybe try this:
```
with tempfile.TemporaryDirectory() as dir:
    wandb.init(...)
    wandb.log(..., dir=dir)
```

nupic/research/frameworks/dendrites/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    num_contexts = context_vectors.size(0)
+    num_dendrites = dendritic_weights.size(0)
+    assert len(mask_values) == num_contexts
+
+    x_labels = [
+        "context {} [{}]".format(j, mask_values[j]) for j in range(num_contexts)
+    ]
+    y_labels = ["dendrite {}".format(j) for j in range(num_dendrites)]
+
+    return wandb.plots.HeatMap(x_labels, y_labels, activations, show_text=False)
```
mvacaporale, line 62: Ideally, we should have our frameworks code independent of `wandb`. Could you use seaborn instead? And then pass that to wandb?

nupic/research/frameworks/dendrites/visualize_demo.py
mvacaporale: Note, `tempfile` is built into python.

mvacaporale (changes requested):

nupic/research/frameworks/dendrites/visualize_demo.py
mvacaporale: Thanks, thankfully `tempfile` is built into python. I included an example below.

nupic/research/frameworks/dendrites/utils.py
subutai: In general that's true for our core codebase. However, I think it's ok for a specific visualization to be dependent on wandb.

mvacaporale: Fair point.

nupic/research/frameworks/dendrites/visualize_demo.py
karangrewal: Ok, you are right about tempfile, but I'm not able to run it without errors on AWS, but locally it's fine. There's another [thread about file permission issues with wandb](https://github.com/wandb/client/issues/1024) which seems unresolved, so half the time it gives an error when I try using a temporary directory with the code sample that you have provided.

Since it's just a demo script, and the function I wrote is the main plus here, I propose we leave the demo script as is for now.

mvacaporale: Fair, I didn't realize there would be issues. Please leave a comment in the docstring just to note to users that a wandb-dir will be made by default, potentially.

mvacaporale (approved):

PR #392, opened 2020-10-23
Introduce the Omniglot dataset via git-lfs (just the zip files)

mvacaporale:
This PR adds the downloaded Omniglot dataset to our testing directory. This will be helpful for more extensive testing of our experiment framework Vernon such as end-to-end tests of the experiment classes.

Supersedes https://github.com/numenta/nupic.research/pull/391

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.pt filter=lfs diff=lfs merge=lfs -text
+tests/unit/frameworks/pytorch/data/** filter=lfs diff=lfs merge=lfs -text
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 *.pt filter=lfs diff=lfs merge=lfs -text
-tests/unit/frameworks/pytorch/data/** filter=lfs diff=lfs merge=lfs -text
+*.zip filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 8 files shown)


.gitattributes
```diff
@@ -1 +1,2 @@
 *.pt filter=lfs diff=lfs merge=lfs -text
+tests/unit/frameworks/pytorch/data/** filter=lfs diff=lfs merge=lfs -text
```
lscheinkman, line 2: To be more generic, I think this could be replaced with:
```
*.zip filter=lfs diff=lfs merge=lfs -text
```

lscheinkman (approved):

subutai:
I'm still not really sure why we need this. In general it is bad practice to include full datasets and other large files in a code focused repo. We've avoided this so far, and I'm not sure I want to start now. Unit tests should run super fast, and thus should only use a few files at most. Anything longer should run in our regression tests, which can separately access datasets.

mvacaporale:
I can see the value in quick unit tests. I'll close this PR and create regression tests instead then. 

@lscheinkman I imagine it will be alright to download the datasets in that case, correct?

lscheinkman:
> I can see the value in quick unit tests. I'll close this PR and create regression tests instead then.
> 
> @lscheinkman I imagine it will be alright to download the datasets in that case, correct?

The regression tests have access to the dataset shared drive. You can just put your dataset there and access from the test. There is no need to download during the test

mvacaporale:
> The regression tests have access to the dataset shared drive. You can just put your dataset there and access from the test. There is no need to download during the test

Thanks, I'll store them there then.

PR #396, opened 2020-10-28
Improved dendrite activation plotting

karangrewal:
Modifications have been made to `plot_dendritic_activations` so that it plots a heatmap of activations where
- the more negative the value, the more red the cell, and
- the more positive the value, the more blue the cell, and

and looks similar to the bottom right plot in [this figure](https://matplotlib.org/3.1.1/_images/sphx_glr_image_annotated_heatmap_003.png).

NOTE: the major change I made is that `plot_dendritic_activations` now uses matplotlib to plot the value, and returns `matplotlib.pyplot` back (line 84) instead of the plot itself, because wandb needs the module in order to plot the image in a wandb run (done as `wandb.log({heatmap: plt})`). Yes, it's strage that we're passing the module back, so I'm curious if there are any better ways to get around this.

### DIFF
--- a/nupic/research/frameworks/dendrites/utils.py
+++ b/nupic/research/frameworks/dendrites/utils.py
@@ -19,8 +19,9 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+import matplotlib.pyplot as plt
+import numpy as np
 import torch
-import wandb
 
 
 def plot_dendritic_activations(
@@ -31,9 +32,9 @@ def plot_dendritic_activations(
 ):
     """
     Returns a heatmap of dendrite activations (given dendritic weights for a single
-    neuron and context vectors) using wandb. Note that the user must be logged in to
-    wandb on both browser and terminal to view the resulting plots, and this can be
-    done via the following command:
+    neuron and context vectors) plotted using matplotlib. Note that the user must be
+    logged in to wandb on both browser and terminal to view the resulting plots, and
+    this can be done via the following command:
 
     $ wandb login your-login-key
 
@@ -44,6 +45,8 @@ def plot_dendritic_activations(
     :param use_absolute_activations: plots absolute activation values if True
     """
     assert dendritic_weights.size(1) == context_vectors.size(1)
+
+    plt.cla()
 
     activations = torch.matmul(dendritic_weights, context_vectors.T)
     if use_absolute_activations:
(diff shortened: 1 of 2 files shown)


nupic/research/frameworks/dendrites/utils.py
```diff
@@ -59,4 +62,23 @@ def plot_dendritic_activations(
...
+    plt.setp(ax.get_xticklabels(), rotation=45, ha="right", rotation_mode="anchor")
+    plt.tight_layout()
+
+    # Annotate just the top absolute activation for each context
+    top_activation_dendrite_per_context = np.argmax(np.abs(activations), axis=0)
+    for j, i in enumerate(top_activation_dendrite_per_context):
+        val = np.round(activations[i, j], 2)
+        ax.text(j, i, val, ha="center", va="center", color="w")
+
+    return plt
```
mvacaporale, line 84: There are several other ways to return what's needed to log your plot to wandb. In fact, wandb accepts many plot-like things. In this case, you could return `fig` or `ax` (I'd recommend the `fig`).

To illustrate the flexibility of what wandb accepts, here are some examples:
```
# %%

import wandb
import matplotlib.pyplot as plt

# Init wandb
wandb.init(name="line_plot", project="test_wandb")

# Give wandb the pyplot module.
plt.plot([1, 2, 3], [1, 2, 3])
plt.title("my_plot")
wandb.log({"plt": plt})

# Give wandb the line object returned.
lines = plt.plot([1, 2, 3], [1, 2, 3])
line = lines[0]  # there's only one in the list
wandb.log({"line": line})  # log line object to wandb

# Give wandb the current figure.
plt.plot([1, 2, 3], [1, 2, 3])
plt.title("my_plot")
fig = plt.gcf()  # get current figure
wandb.log({"fig": fig})


# Given wandb the current axis.
plt.plot([1, 2, 3], [1, 2, 3])
plt.title("my_plot")
ax = plt.gca()  # get current axis
wandb.log({"ax": ax})
```

Note, I was somewhat confused why one could use either `fig` or `ax`. How can these be treated by `wandb` as the same thing? This [link](https://realpython.com/python-matplotlib-guide/#:~:text=A%20Figure%20object%20is%20the,%E2%80%9D%20as%20we%20might%20expect) helped. It seems `ax` is actually a nested object within `fig`.  In the end, wandb just wants the figure, but it can get it by multiple means. It can access `ax.figure` or it can ask the module for the current figure as in `plt.gcf()`. Here's a distilled version of their logic:
```
def get_matplotlib_figure(obj):
    if obj == matplotlib.pyplot:
        obj = obj.gcf()
    elif not isinstance(obj, Figure):
        if hasattr(obj, "figure"):
            obj = obj.figure
       else:
            raise ValueError
    return obj
```

mvacaporale (approved):

nupic/research/frameworks/dendrites/utils.py
karangrewal: Thanks for looking into this, I decided to go with `plt.gcf()` and return that value.

mvacaporale: That works. Although I believe you could have skipped the call to `plt.gcf()` and returned `fig` as it's defined earlier in the function.

PR #399, opened 2020-10-30
Fix accidental installs of pytorch 1.7

mrcslws:


mrcslws:
Interesting @lscheinkman 

Before I added the fastai version number, it did:

```
Collecting torch>=1.6
  Downloading torch-1.7.0-cp37-cp37m-manylinux1_x86_64.whl (776.7 MB)
```

With the fastai version number, it does:

```
Collecting torch<1.7,>=1.6
  Downloading torch-1.6.0-cp37-cp37m-manylinux1_x86_64.whl (748.8 MB)
```

(The >=1.6 dependency comes from gpytorch.)

So the problem is that nupic.torch's dependency is not incorporated into this pip decision.

lscheinkman (approved):
The new pip dependency resolver is very confusing

PR #401, opened 2020-11-01
Anchor colorbar (in dendrite activation heatmap)

karangrewal:
Using matplotlib, the heatmap of dendrite activations is now anchored such that all negative values are red, and all positive values are blue (a value of zero will be colored white). This will help with visualizations.

[Here](https://drive.google.com/file/d/14F8GqMJZQXCyagM2oSa7u0T5T59eCFQd/view?usp=sharing) is a concrete example -- this follows the "negative=red" and "positive=blue" rule, and although +10 is the largest positive value, it's not dark blue since the largest negative value is -19.


lucasosouza (approved):
Awesome!

PR #403, opened 2020-11-04
Simplify VaryBatchSize mixin, remove unneeded self.batch_size

mrcslws:
I'm working on adding an interface that allows the VaryBatchSize mixin to adjust the `self.total_steps` property so that it is compatible with OneCycleLR. That change will be easier with this simplified VaryBatchSize code.

Note that it's better to do `isinstance(v, Sequence)` instead of `isinstance(v, list)`, because tuples should also be allowed.

### DIFF
--- a/nupic/research/frameworks/vernon/mixins/vary_batch_size.py
+++ b/nupic/research/frameworks/vernon/mixins/vary_batch_size.py
@@ -19,7 +19,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-from copy import deepcopy
+from collections.abc import Sequence
 
 
 class VaryBatchSize(object):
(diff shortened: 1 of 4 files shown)


nupic/research/frameworks/vernon/mixins/vary_batch_size.py
```diff
@@ -33,66 +33,39 @@ class VaryBatchSize(object):
...
 
     def setup_experiment(self, config):
-
-        super().setup_experiment(config)
-
         # Get and validate batch sizes.
         batch_sizes = config.get("batch_sizes", None)
-        assert isinstance(batch_sizes, list), "Must specify list of batch sizes"
-        self.batch_sizes = batch_sizes
+        assert isinstance(batch_sizes, Sequence), "Must specify list of batch sizes"
```
mvacaporale, line 38: The comment should now read "Must specify a sequence of batch sizes"

nupic/research/frameworks/vernon/mixins/vary_batch_size.py
```diff
@@ -33,66 +33,39 @@ class VaryBatchSize(object):
...
-
-        # Set the batch size to the desired variant.
-        batch_size = batch_sizes[batch_size_variant]
-        config = deepcopy(config)
-        config.update(batch_size=batch_size)
+        super().pre_epoch()
+        if 0 < self.current_epoch < len(self.batch_sizes):
+            batch_size = self.batch_sizes[self.current_epoch]
+            self.train_loader = self.create_train_dataloader(
+                {**self.config, "batch_size": batch_size}
```
mvacaporale, line 58: Cool trick to know.

nupic/research/frameworks/vernon/mixins/vary_batch_size.py
mrcslws: Ok, I can change it. (Often in code comments I use "list" in the generic English sense. Usually I rely on duck-typing rather than type checking -- as long as they pass in something that supports indexing and len, the code will work)

mvacaporale: I see. That's fair. Either way.

PR #406, opened 2020-11-10
Absolute max convolutional dendrite layer

karangrewal:
The new class `AbsoluteMaxGatingDendriticLayer2d` implements a convolutional layer using dendrite activations. The dendrite activations are applied to each channel (i.e., all activations in a single output channel from the conv layer are multiplied by a single value computed by the dendrites). Test cases are included.

**Update 11/15:** I also added `GatingDendriticLayer2d` along with unit tests.


nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -107,3 +108,72 @@ def apply_dendrites(self, y, dendrite_activations):
...
+
+        self.rezero_weights()
+
+    def rezero_weights(self):
+        super().rezero_weights()
+        if self.segments is not None:  # only none at beggining of init
+            self.segments.rezero_weights()
+
+    def apply_dendrites(self, y, dendrite_activations):
+        output_max = dendrite_activations.max(dim=2).values
```
mvacaporale, line 153: I wonder if this can be done in a simpler way. Possibly, you can call `idices = dendrite_activations.abs().max(dim=2).indices` and then unpack the indices as appropriate.

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -107,3 +108,72 @@ def apply_dendrites(self, y, dendrite_activations):
...
+        )
+
+        self.rezero_weights()
+
+    def rezero_weights(self):
+        super().rezero_weights()
+        if self.segments is not None:  # only none at beggining of init
+            self.segments.rezero_weights()
+
+    def apply_dendrites(self, y, dendrite_activations):
```
mvacaporale, line 152: Could you please add a docstring and could you elaborate on the computation being done below?

karangrewal: Done.

karangrewal:
@mvacaporale I just added `GatingDendriticLayer2d` (which is a simple extension of `AbsoluteMaxGatingDendriticLayer2d`). I'm currently looking into how our absolute max implementation can be improved as you suggested.

nupic/research/frameworks/dendrites/dendritic_layers.py
karangrewal: Yes, just made the change!

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -101,13 +102,111 @@ class AbsoluteMaxGatingDendriticLayer(BiasingDendriticLayer):
...
+        """
```
subutai, line 153: Shouldn't there be a forward() function that calls apply_dendrites? Then you can do `y = dendrite_layer(x, context)`.

karangrewal: Yes, you're right. I just added `forward` (and copied from the 1d dendrite layers).

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -102,16 +102,12 @@ class AbsoluteMaxGatingDendriticLayer(BiasingDendriticLayer):
...
-        )
-        return y * torch.sigmoid(
-            torch.abs(dendrite_activations).max(dim=2).values * sign_mask
-        )
+        inds = dendrite_activations.abs().max(dim=2).indices
+        inds = inds.unsqueeze(dim=2)
+        dendrite_activations = torch.gather(dendrite_activations, dim=2, index=inds)
+        dendrite_activations = dendrite_activations.squeeze()
+        dendrite_activations = torch.sigmoid(dendrite_activations)
+        return y * dendrite_activations
```
mvacaporale, line 110: This looks good, definitely. Did you time this method to compare it to the previous one?

mvacaporale (approved):

nupic/research/frameworks/dendrites/dendritic_layers.py
karangrewal: I didn't, but I'll do that (probably after I merge) and will get back to you about it.

mvacaporale: Thanks, please do. It's always good to do a sanity check on these sorts of things. Indices especially have many ways to use them, and they each could have their own tradeoffs with respect to time and memory resources.

subutai (changes requested):
Shouldn't the tests cover using the forward() function as well? After that, feel free to merge.

tests/unit/frameworks/dendrites/dendrites_test.py
```diff
@@ -457,5 +457,537 @@ def test_gradients(self):
...
+            ]
+        ])
+
+        # Context input to dendrite layer: batch_size=2, dim_context=4
+        context_vectors = torch.tensor([
+            [1.0, 0.0, 1.0, 0.0],
+            [0.0, 1.0, 0.0, 0.0]
+        ])
+
+        # Expected absolute max dendrite activations (pre-sigmoid):
```
subutai, line 819: Why is this commented out?

subutai (approved):

tests/unit/frameworks/dendrites/dendrites_test.py
karangrewal: All these comments are just to show step-by-step how we arrive at the hand-derived solution which the output from `forward should match`.

subutai: Oh got it, that's quite useful.

PR #407, opened 2020-11-10
Updates to MetaCL Experiment Class and OML Configs

mvacaporale:
This includes fixes, configs to closer replicate the OML repo, and some refactoring to distinguish the fast and slow params through the config.

Fixes include:
* Setting the epochs for the dataloaders when distributed; otherwise they constantly use the same seed and sample the same data.
* Move `pre_task` prior to cloning the model; otherwise the `rest_task_params` has no influence on meta-training.

Configs include:
* A config for a 2000 epoch model, just for quick regression test purposes. Eventually, this can be placed with the actual regression tests.
* Configs that replicate OML's repo more closely. See `oml_replicate.py`.

Refactoring includes:
* We can now determine which params serve what purpose via the config. This looks like the following
```
config = dict(
    ...,

    # Params to train in inner loop - given as regex pattern.
    fast_params=["adaptation.*"],

    # Params to train during meta-test training.
    test_train_params=["adaptation.*"],

    # Params to reset prior to training a task.
    # These will also be re-initialized prior to each meta-testing run.
    output_layer_params=["adaptation.0.weight", "adaptation.0.bias"],
)
```
^ Slow params are not included as the MetaCL experiment automatically trains all of them.

This params-via-config will be helpful when we create various types of dendrite models. We can define the model however we want, and then adjust the config appropriately. For instance, we may set `fast_params=[features.segments*]`, or whatever seems reasonable (we could just as easily train them in the outer loop).


nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
```diff
@@ -118,6 +123,12 @@ def setup_experiment(self, config):
         self.replay_batch_size = config.get("replay_batch_size", 64)
         self.num_fast_steps = config.get("num_fast_steps", 1)
 
+        assert "fast_params" in config
+        fast_named_params = filter_params(self.model,
+                                          include_patterns=config["fast_params"])
+        self.fast_param_names = list(fast_named_params.keys())
```
lucasosouza, line 129: Can't we just return the keys directly, since the params are not being used at all?

nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
```diff
@@ -160,7 +171,7 @@ def create_train_slow_sampler(cls, config, dataset):
     @classmethod
     def create_replay_sampler(cls, config, dataset):
         """Sampler used to augment meta-train testing; "replays" previous classes."""
-        class_indices = cls.compute_class_indices(config, dataset, mode="replay")
+        class_indices = cls.compute_class_indices(config, dataset, mode="all")
```
lucasosouza, line 174: What about that distinction we discussed, of having the replay only use half of the classes? How did you set that?

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -56,21 +65,44 @@ def setup_experiment(self, config):
...
+        self.num_meta_test_classes = config.get("num_meta_test_classes",
+                                                [10, 50, 100, 200, 600])
+
+        # Resolve the names of the meta-test training params.
+        assert "test_train_params" in config
+        test_train_named_params = filter_params(
+            self.model,
+            include_patterns=config["test_train_params"]
+        )
+        self.test_train_param_names = list(test_train_named_params.keys())
```
lucasosouza, line 90: Same comment as above, we never end up using the params.

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -123,21 +155,53 @@ def create_test_test_dataloader(cls, config, dataset):
             pin_memory=torch.cuda.is_available(),
         )
 
+    def pre_task(self, tasks):
+        """Re-initialize task params prior to meta-train training."""
+        super().pre_task(tasks)
+        if not self.reset_task_params:
```
lucasosouza, line 161: Suggestion to to do `if self.reset_task_params: <code>`. There is no need for the return here since this function isn't expected to return anything.

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -123,21 +155,53 @@ def create_test_test_dataloader(cls, config, dataset):
...
+                    task_weights = p[t, :].unsqueeze(0)
+                    nn.init.kaiming_normal_(task_weights)
+
     def post_epoch(self):
         super().post_epoch()
-        if self.should_stop() and self.run_meta_test:
-            # meta-training phase complete, perform meta-testing phase
-            for num_classes_learned in [10, 50, 100, 200, 600]:
-                if num_classes_learned > self.num_classes_eval:
+        if self.current_epoch >= self.epochs - 1 and self.run_meta_test:
```
lucasosouza, line 173: If self.current_epoch is zero indexed, isn't self.epochs -1 the last epoch? If that is the case == makes more sense than >=

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -123,21 +155,53 @@ def create_test_test_dataloader(cls, config, dataset):
...
+
+            # Accumulate results.
+            table = []
+            headers = ["Num Classes", "Meta-test test", "Meta-test train", "LR"]
+
+            # Meta-training phase complete, perform meta-testing phase
+            for num_classes in self.num_meta_test_classes:
+                if num_classes > self.num_classes_eval:
```
lucasosouza, line 181: I didn't get the meaning behind this if statement, a comment would help

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -123,21 +155,53 @@ def create_test_test_dataloader(cls, config, dataset):
...
-                test_train_accs, test_test_accs = results
-                print("Accuracy for meta-testing phase over"
-                      f" {num_classes_learned} num classes.")
-                print("Meta-test training accuracies:", test_train_accs)
-                print("Meta-test tests accuracies:", test_test_accs)
+                # TODO: log results in addition to simply printing them to stdout
+                results = self.run_meta_testing_phase(num_classes)
+                test_train_accs, test_test_accs, lr = results
+
+                print(f"Accuracy for meta-testing phase over {num_classes} num classes")
```
lucasosouza, line 188: We can probably refactor this later to either hide it behind a verbose flag or use a higher level of logging, so we can decide whether or not to print the table.

But I understand now it is the only output, so it is required.

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -152,7 +216,7 @@ def find_best_lr(self, num_classes_learned):
         lr_all = []
 
         # Grid search over lr
-        for _lr_search_runs in range(0, self.num_lr_search_runs):
+        for _lr_search_run in range(0, self.num_lr_search_runs):
```
lucasosouza, line 219: The _ is to signal this variable is not being used? I would rather use just a regular underscore, I found myself looking for it to understand the meaning of this variable.

nupic/research/frameworks/vernon/mixins/oml.py
```diff
@@ -266,7 +329,27 @@ def run_meta_testing_phase(self, num_classes_learned):
             acc = correct / len(self.test_train_loader.sampler.indices)
             meta_test_train_accuracies.append(acc)
 
-        return meta_test_train_accuracies, meta_test_test_accuracies
+        return meta_test_train_accuracies, meta_test_test_accuracies, lr
+
+    def get_named_test_train_params(self):
```
lucasosouza, line 334: There is only one variable difference between these 2 functions, can't they just be a `get_named_params` functions with an additional argument?

lucasosouza, line 334: There is only a one line difference between these 2 functions, can't they just be a `get_named_params` function with an additional argument/flag?

lucasosouza (approved):
Looks good, left some non-essential code style comments.

nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
mvacaporale: We could, but the filter_params is intended to be more general. We use just the names, but it could be helpful to retrieve the params to reinitialize them, for instance.

nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
mvacaporale: There's a subclassed version of the experiment in `meta_cl_data_split.py` in the project folder that does it.

I figure we should actually use all of the data, so I didn't incorporate that option into the main codebase. We can add it in anytime if we'd like, maybe a three-line change.

nupic/research/frameworks/vernon/mixins/oml.py
mvacaporale: The `self.num_classes_eval` are the number of meta-testing classes. This prevents sampling, say, 700 classes from a meta-test set comprised of only 659 classes.

nupic/research/frameworks/vernon/mixins/oml.py
mvacaporale: Changed it

nupic/research/frameworks/vernon/mixins/oml.py
mvacaporale: I made a new function called `_get_params_by_names` that simplifies each of these to one line, as well as `get_named_fast_params`.

nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
lucasosouza: But are we using it for that? It sounds a bit like overengineering for some future requirement but may not even happen.

nupic/research/frameworks/vernon/experiments/meta_cl_experiment.py
lucasosouza: I think it is good idea to add it. If we want to compare our results to OML for the paper, we should be able to easily reproduce their results within the framework.

mvacaporale: Fair enough. Can do.

mvacaporale: There are now two params `slowfast_classes` and `replay_classes` that be passed to the config to list which classes which will be used for which dataset. To replicate OML, one would set `replay_classes=range(0, 481)` and `slowfast_classes=range(481, 963)`.

lucasosouza: Sounds cool. We should think about the `slowfast` naming in the future, I agree with Subutai that sounds a bit off. Maybe just `train_classes`.

karangrewal (approved):
Something to consider: replace "slowfast" with another name that's more intuitive.

mvacaporale:
> Something to consider: replace "slowfast" with another name that's more intuitive.

Good point, I changed it to `fast_and_slow_classes`.

PR #417, opened 2020-11-24
Update dendritic network and config

mvacaporale:
This introduces `metacl_dendrites2` config. It performs much worse than `metacl_dendrites` (can not train as of yet), but it's closer in network architecture and training paradigm to ANML.

There's also a new debugging notebook: `nupic.research/projects/meta_cl/notebooks/debug_metacl_experiment.ipynb`. The comments hopefully make it clear how to use it. Essentially, the idea is to achieve quick development cycles on plots and figures that help debug any meta-cl experiment. You run each epoch manually, and so you can easily add plots mid-way through training as desired. 

This PR depends on `https://github.com/numenta/nupic.research/pull/416/files` to rezero the weights properly.


projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -107,3 +109,107 @@ def forward(self, x):
         pred = self.prediction(x)
         out = self.gating_layer(pred, context=mod)
         return out
+
+
+class DendriticNetwork2(nn.Module):
```
karangrewal, line 114: Maybe give this a better name, since we will probably go through many iterations of network architecture. Although, there's an argument to be made for this naming style too (e.g., iPhone3, iPhone4, iPhone5, etc). Do you think we can make it more specific?

nupic/research/frameworks/vernon/mixins/rezero_weights.py
```diff
@@ -49,7 +49,7 @@ def create_model(cls, config, device):
         # Some initialization strategies can destroy sparsity, so we call rezero
         # here.
         for module in model.modules():
-            if isinstance(module, SparseWeightsBase):
+            if isinstance(module, HasRezeroWeights):
```
karangrewal, line 52: (minor comment) I see you made this same change in #416, any reason you're re-doing it here?

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -107,3 +109,107 @@ def forward(self, x):
...
+        dendrite_activations = dendrite_activations.squeeze(dim=2)
+        dendrite_activations = torch.sigmoid(dendrite_activations)
+        return y * dendrite_activations
+
+    def forward(self, x):
+        mod = self.modulation(x)
+        pred = self.prediction(x)
+
+        dendrite_activations = self.segments(mod)
+        out = self.apply_dendrites(pred, dendrite_activations)
```
karangrewal, line 211: why not just use the `AbsoluteMaxGatingDendriticLayer` here, instead of redefining the `forward` and `apply_dendrites` methods?

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -107,3 +109,107 @@ def forward(self, x):
...
+
+    With default parameters and `num_classes-963`, it uses 5,118,291 weights-on
+    out of a total of 8,345,095 weights. In comparison, ANML uses 5,963,139 weights
+    and OML uses 5,172,675 weights. Thus, these is still room to add more.
+    """
+
+    def __init__(self, num_classes,
+                 num_segments=20,
+                 dim_context=100,
+                 dendrite_sparsity=0.70,
```
karangrewal, line 133: In case the performance of this model is really bad, I think it may be because if 70% of dendrite weights are initialized as zero, than they won't get updated in the backward pass, so only 30% of dendrite weights can be useful during the lifetime of the network. That's just a guess - what do you think?

karangrewal (commented):
Here are my comments.

projects/meta_cl/networks/dendritic_networks.py
karangrewal: Actually what I stated above may be incorrect, since the derivative w.r.t. the dendrite weight just involves the dendrite activation value, and not the zero weight.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: Fair. We can call it `ANMLDendriticNetwork`, since it's mimicking theirs.

nupic/research/frameworks/vernon/mixins/rezero_weights.py
mvacaporale: When I made this PR, the other one (#416) hadn't been merged yet, but I built off of its changes. You'll see that once this PR is up to date with the master branch that these diffs won't be highlighted anymore.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: It's actually a different set of operations being performed here. The gating is being applied to the flattened output of a conv layer, and `AbsoluteMaxGatingDendriticLayer` doesn't do that nor does its 2d counterpart.

When we made the `DendriticSegments`, I had it in mind to be able to use it directly as needed whenever we didn't have a layer that neatly wrapped the desired functionality. As we play around with architectures, I'm sure we'll find some canonical form and we can make higher-level classes accordingly.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: You're right, the derivative may not be zero. This has a complicated story though.

I meant to keep the weights always zero. The module `DendriteSegments` is supposed to behave very similarly to `SparseWeights`. Both inherit from `HasRezeroWeight` so they both have a `rezero_weights` method that should be called after every gradient step.

However, I just realized I fell short of this. I intended to have the `RezeroWeights` mixin take care of rezero-ing the weights (in PR #416, this mixin now operates over every `HasRezeroWeights` module`), but I did not realize it does so in the `post_batch` method, which `MetaContinualLearningExperiment` doesn't have.

Here's a Jira to accomplish this https://ntaresearch.atlassian.net/browse/RES-1861. I'll take care of this soon, either today or tomorrow so we know the modules are behaving as we want them to.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: I also changed `metacl_dendrites2` to `metacl_anml_dendrites`.

karangrewal (approved):

projects/meta_cl/experiments/dendrites.py
```diff
@@ -20,11 +20,23 @@
...
 
-from networks import DendriticNetwork
+from networks import ANMLDendriticNetwork, DendriticNetwork
+from nupic.research.frameworks.vernon import MetaContinualLearningExperiment, mixins
 
+from .anml_replicate import ANMLTransform, metacl_anml_replicate
 from .oml_replicate import metacl_oml_replicate
 
 
+class DendritesExperiment(mixins.RezeroWeights,
```
lucasosouza, line 30: I am curious of the performance with and without rezeroweights (dense vs sparse). I saw in the comment below it is not really sparse now because it never calls the `post_batch` method, but at some point later we can evaluate.

lucasosouza (approved):

projects/meta_cl/experiments/dendrites.py
mvacaporale: Good point. As of now, it doesn't get past 0 accuracy (both meta-training and testing), so we'll need a working network for comparison.

mvacaporale: Maybe that's what's making it harder to train now. I wouldn't doubt if it's too over parameterized without the sparse weights.

PR #424, opened 2020-12-19
Simplify `train_model` and `evaluate_model`

karangrewal:
The parameter `freeze_params` has been removed as it was unused, and `active_classes` has been added. Overall, this simplifies both the train and evaluate functions since the removed parameter consumed much more lines of code than the newly-introduced parameter.

Note:
The implementation for `active classes` borrows from [here](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/continual_learning/multihead/multihead.py).

### DIFF
--- a/nupic/research/frameworks/pytorch/model_utils.py
+++ b/nupic/research/frameworks/pytorch/model_utils.py
@@ -142,6 +142,7 @@ def train_model(
                                                               t4 - t3, t5 - t4)
             post_batch_callback(model=model,
                                 error_loss=error_loss.detach(),
+                                complexity_loss=None,
                                 batch_idx=batch_idx,
                                 num_images=num_images,
                                 time_string=time_string)
--- a/nupic/research/frameworks/pytorch/model_utils.py
+++ b/nupic/research/frameworks/pytorch/model_utils.py
@@ -141,7 +141,7 @@ def train_model(
                            + "weight update: {:.3f}s").format(t1 - t0, t2 - t1, t3 - t2,
                                                               t4 - t3, t5 - t4)
             post_batch_callback(model=model,
-                                error_loss=error_loss.detach()
+                                error_loss=error_loss.detach(),
                                 batch_idx=batch_idx,
                                 num_images=num_images,
                                 time_string=time_string)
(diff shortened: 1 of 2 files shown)


mrcslws:
A couple thoughts:
- `transform_to_device_fn` is needed for knowledge distillation. So, it's used by most of our imagenet experiments. 
- Our highest accuracy imagenet experiments use complexity loss
- In case you're wondering why error loss and complexity loss are separate: this is partly because both can be memory intensive, and there's no reason they need to fight over GPU memory.

Are you suggesting doing two different optimizer steps, first the error_loss step then the regularization step? Maybe with the regularization step in a post_batch_callback? It's better to perform a single step that incorporates the entire loss function, especially when using momentum. So the complexity loss needs to be calculated before the optimizer step.

If we want a simple train_model function that's not capable of knowledge distillation or complexity loss, maybe it should be a separate function?

Here's one bigger possible change that could address all of this, inspired by Lightning. In Lightning, each model specifies its "training step". For example: https://github.com/numenta/nupic.research/blob/41a98a3c24481bb5d1615c879abc8ae268580685/nupic/research/frameworks/lightning/models/supervised_model.py#L71

`train_model` could take a `training_step_fn` that defaults to a built-in simple `supervised_training_step` function. Custom functions could be plugged in to use knowledge distillation, complexity loss, or any other experiment-specific code.

mrcslws:
Note also that the `l1_regularization_step` function you linked to won't work properly with momentum. https://github.com/numenta/nupic.research/blob/41a98a3c24481bb5d1615c879abc8ae268580685/nupic/research/frameworks/pytorch/regularization.py#L26

If the error gradient has lower magnitude than the L1 gradient, it will be able to build up momentum and "outrun" the L1 gradient. That's not possible when implementing L1 regularization via the loss function. L1 regularization should modify the tensors' gradients, it shouldn't be treated as a separate step. (Though that approach does work without momentum.)

lucasosouza:
Reiterating that filtering active classes when calculating the loss won't solve for scenarios where we have a metric dependent on the entire network or layer, such as weight norm, a complexity loss on the weights of the entire model or layer (see ewc), sampling for dropconnect or dynamic sparsity.

It might be a good intermediate solution to get things going. But if we are making big changes, we should aim for a more definitive solution.

mrcslws (approved):

PR #426, opened 2020-12-23
Updates to dendrite plotting function

karangrewal:
The function `plot_dendrite_activations` has been modified so that

(a) parameter `use_absolute_activations` has been removed, as this was unused anyways,
(b) parameter `annotation_type` has been added as this controls whether the maximum or absolute maximum activation per context get labeled in the plots (depending on the type of dendrite gating layer), and
(c) parameter `mask_values` is now None by default and does not need to be used, since this is only useful for the dendrite routing test.


lucasosouza (approved):
Changes look good to me, but I am not familiar with this function.

PR #427, opened 2020-12-30
Minor update to metacl experiments and results

mvacaporale:
This includes a variety of relatively small updates:
* Added configs to replicate ANML's results, but with a dendritic network (see comment  below), 
* Added network in `dendritic_networks.py` to help with the above bullet.
* The `ANMLsOMLNetwork` has some changes to make it's archtecture a more accurate recreation.
* Some typo fixes and other minor enhancements.

Two config's helps recreate ANMLs results with a dendritic network.
1. `metacl_anml_dendrites_adjust_lr`: This adjusts the lr to ensure it's using the same as ANML's. It was a mistake not to do so previously.
2. `closer_to_anml_dendrites`: This uses the adjusted lr, as well an adjusted network that makes our dendritic network identical to ANMLs. (i.e it uses num_segments=1).

Both of these changes incrementally improves the accuracy. Here are the meta-testing results from ANML ran in Vernon:
|   Num Classes | Meta-test test   | Meta-test train   |    LR |
|--------------:|:-----------------|:------------------|------:|
|            10 | 0.88 ± 0.05      | 0.94 ± 0.03       | 0.001 |
|            50 | 0.88 ± 0.02      | 0.98 ± 0.01       | 0.001 |
|           100 | 0.85 ± 0.02      | 0.98 ± 0.01       | 0.001 |
|           200 | 0.80 ± 0.01      | 0.97 ± 0.01       | 0.001 |
|           600 | 0.66 ± 0.01      | 0.93 ± 0.00       | 0.001 |

Here are the meta-testing results of `closer_to_anml_dendrites`:
|   Num Classes | Meta-test test   | Meta-test train   |    LR |
|--------------:|:-----------------|:------------------|------:|
|            10 | 0.86 ± 0.05      | 0.92 ± 0.03       | 0.001 |
|            50 | 0.89 ± 0.03      | 0.98 ± 0.01       | 0.001 |
|           100 | 0.86 ± 0.02      | 0.98 ± 0.01       | 0.001 |
|           200 | 0.80 ± 0.01      | 0.97 ± 0.01       | 0.001 |
|           600 | 0.68 ± 0.01      | 0.93 ± 0.00       | 0.001 |

These results provide a sanity check to ensure our dendrite network can train as expected when `num_segments=1`.


projects/meta_cl/experiments/dendrites.py
```diff
@@ -105,17 +109,21 @@ def run_experiment(config):
 
 
 # An updated version of the dendrites model.
-# |--------------------------------------------------------------|
-# |   Num Classes | Meta-test test   | Meta-test train   |    LR |
-# |--------------:|:-----------------|:------------------|------:|
-# |            10 | 0.00 ± 0.00      | 0.00 ± 0.00       | 0.001 |
-# |--------------------------------------------------------------|
+# |---------------------------------------------------------------|
```
lucasosouza, line 112: We quickly discussed about this, but it is significantly faster if we don't gridsearch LR during test time. We can still do that for final results, but I don't clearly see the benefits of doing that every run during experimentation.

It can be argued that there might be better LRs for each method/architecture, but that can be said for any hyperparameter. It is hard to put the finger on what is a good trade-off during experimentation. But I do feel that meta-testing as it is takes a long time and its slowness is hindering progress (it did for me at least, I turned off when running experiments).

A second argument, but more opinionated, is that I don't believe it fits the concept of continual learning to do hyperparameter search during test time, specially given the test data distribution is similar to the one seen during meta-training. There is no application scenario I can imagine that fits this odd paradigm, this is purely an artifact of ANML/OML starting from a copy of the earlier MAML implementation. The optimal LR should be picked up as part of the meta learning process.

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -184,6 +182,116 @@ def conv_block(cls, in_channels, out_channels, kernel_size, stride, padding):
...
+
+        return out
+
+
+class CloserToANMLDendriticNetwork(nn.Module):
+    """
+    An iteration of a dendritic network, based closely on `ANML`_. (The conv + max-pool
+    layers are identical). It's similar to ANMLDendriticNetwork, but does not include a
+    linear layer atop the modulation network. This makes it more similar to ANML's
+    original network. In fact, with `num_segments=1, dendrite_sparsity=0`, they should
```
lucasosouza, line 223: Is it correct to conclude that they are? Besides the empirical results with the full training, did you also look at the networks via other means and came to that same conclusion?

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -184,6 +182,116 @@ def conv_block(cls, in_channels, out_channels, kernel_size, stride, padding):
...
+    be identical.
+
+    .. _ANML: nupic.research/projects/meta_cl/networks/anml_networks.py
+
+    Input shape expected is 3 x 28 x 28 (C x W x H) but an adaptive average pooling
+    layer helps accept other heights and width.
+
+    With default parameters and `num_classes-963`, it uses 5,132,832 weights-on
```
lucasosouza, line 231: Just double checking, is the ANML network output size 963, exact number of classes as in the background set? I had it in my head it was 1000.

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -184,6 +182,116 @@ def conv_block(cls, in_channels, out_channels, kernel_size, stride, padding):
...
+    out of a total of 13,960,323 weights. In comparison, ANML uses 5,963,139 weights
+    and OML uses 5,172,675 weights. Thus, these is still room to add more.
+
+    The default parameters are chosen so that the number of on weights for this network
+    is close to that of ANMLDendriticNetwork.
+    """
+
+    def __init__(self, num_classes,
+                 num_segments=10,
```
lucasosouza, line 240: If it mimics ANML is the goal, shouldn't the default values be num_segments=1 and dendrite_sparsity=0, as mentioned in the docstring?

projects/meta_cl/networks/dendritic_networks.py
```diff
@@ -184,6 +182,116 @@ def conv_block(cls, in_channels, out_channels, kernel_size, stride, padding):
...
+            nn.MaxPool2d(kernel_size=2, stride=2),
+            *self.conv_block(112, 112, 3, 1, 0),
+            nn.MaxPool2d(kernel_size=2, stride=2),
+            *self.conv_block(112, 112, 3, 1, 0),
+            nn.AdaptiveAvgPool2d(output_size=(3, 3)),
+            nn.Flatten(),
+        )
+
+        # Apply Kaiming initialization
+        self.reset_params()
```
lucasosouza, line 276: Not sure this is necessary. I believe Kaiming is the default initialization for Linear and Conv2d. (https://discuss.pytorch.org/t/whats-the-default-initialization-methods-for-layers/3157/21)

lucasosouza (approved):
Great results! Have some comments and questions, nothing too relevant.

projects/meta_cl/experiments/dendrites.py
mvacaporale: All good points. I can include an option to skip it. I'd like to keep it to enable fair comparisons to OML and ANML.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: I did not. I didn't want to spend too much time on this considering it was just a sanity check. Manual inspection shows that the architectures are identical baring the linear layers swapped for the dendrites layer, so I left it there.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: The ANML repo used 1000 but we used 963. In my experiments, this didn't seem to make a difference, so I left it there.

projects/meta_cl/networks/dendritic_networks.py
mvacaporale: That was a subgoal. Another goal was to create a network similar to ANML but one that had multiple segments. The network serves both of those goals.

projects/meta_cl/networks/dendritic_networks.py
lucasosouza: Sure, I just thought it was important to mention since you were tracking down the last few points in difference in accuracy between the two repos.

PR #434, opened 2021-01-11
RES-1896: Update GSC experiment making it compatible with original GSC experiment

lscheinkman:
- Add option to evaluated noise 
- Add option to run multiple seeds
- Test GSC "test" dataset instead of "validation" dataset


nupic/research/frameworks/pytorch/datasets/gsc_factory.py
```diff
@@ -57,9 +59,18 @@ def preprocessed_gsc(root, train=True, download=True):
     if train:
         basename = "gsc_train"
         qualifiers = range(30)
```
subutai, line 61: Can you make the range a parameter, defaulting to 30? We have 100 in our internal version.

nupic/research/frameworks/pytorch/datasets/gsc_factory.py
```diff
@@ -46,7 +46,9 @@ def preprocessed_gsc(root, train=True, download=True):
     will be reused.
 
     :param root: directory to store or load downloaded data
-    :param train: whether to load train of validation data
+    :param train: whether to load train or test data
```
subutai, line 49: Having 3 boolean parameters here is really confusing. In addition, to test on the test set you have to set all three to False, which is quite strange. 

There are 8 possible choices - are all valid? (Clearly not, there are only 4 possible outcomes.) There is no need to maintain backward compatibility here, so feel free to use a more natural version.

projects/gsc/mixins.py
```diff
@@ -0,0 +1,86 @@
...
+        eo = super().get_execution_order()
+        eo["setup_experiment"].append("GSCNoiseTest")
+        eo["create_loaders"].append("GSCNoiseTest")
+        eo["validate"].append("GSCNoiseTest")
+
+    def validate(self, loader=None):
+        """
+        Validate noise
+        """
+        results = super().validate(loader=loader)
```
subutai, line 68: This is likely to be a performance hit. Can we do this only for specific epochs?

subutai (changes requested):
Looks good! Do the existing experiments replicate all our whydense results in  nupic.research/projects/whydense/gsc/experiments.cfg? If not, could you add them in?

projects/gsc/mixins.py
lscheinkman: This method will only be called on the `epochs_to_validate` epochs. Do you need a separate one for noise?

subutai: No, that's perfect.

PR #436, opened 2021-01-13
Introduce jupyter notebook to compute sparsity stats per layer

mvacaporale:
This is the jupyter notebook I presented at today's meeting (9/13/2021).

Depending on how this best serves the team, this is not the final version. If either of you have time over the next few days, please try it out and let me know what you think (Note: It should work for any type of model, supervised, metacl, ect., just change the params at the top).

These metrics could be automated to measure throughout training or just at the end of training, with the former taking a bit more work. If there's interest in this, please let me know.


lucasosouza (approved):
Great, thanks

karangrewal (approved):

PR #441, opened 2021-01-25
RES-1923: Refactored dendrite layer classes to extend from base classes

alexcuozzo-1:


nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -41,8 +40,8 @@ class BiasingDendriticLayer(SparseWeights):
     """
 
     def __init__(
-        self, module, num_segments, dim_context,
-        module_sparsity, dendrite_sparsity, dendrite_bias=None
+            self, module, num_segments, dim_context,
+            module_sparsity, dendrite_sparsity, dendrite_bias=None
```
karangrewal, line 44: Why did you indent the list of parameters by an extra 4 spaces?

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -104,17 +108,18 @@ def apply_dendrites(self, y, dendrite_activations):
...
+    Similar to the DendriticLayerBase class, the output from the dendrite segments
+    is applied to the output of each channel. Thus, each channel output gets
+    modulated by a set of dendritic segments.
     """
 
     def __init__(
-        self, module, num_segments, dim_context,
-        module_sparsity, dendrite_sparsity, dendrite_bias=None
+            self, module, num_segments, dim_context,
+            module_sparsity, dendrite_sparsity, dendrite_bias=None
```
karangrewal, line 122: Extra indent again.

karangrewal (approved):
Looks good, but please fix the indents that I pointed out.

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -68,31 +67,36 @@ def __init__(
         self.rezero_weights()
 
     def rezero_weights(self):
+        """Set the previously selected weights to zero."""
         super().rezero_weights()
         if self.segments is not None:  # only none at beginning of init
             self.segments.rezero_weights()
 
     def apply_dendrites(self, y, dendrite_activations):
```
mvacaporale, line 75: You should add the `@abc.abstractmethod` decorator (here is an [example](https://github.com/numenta/nupic.torch/blob/0f0e2452dd648ef0421d53d872c5aaa4b6a7feae/nupic/torch/modules/sparse_weights.py#L63)). This ensures that the `apply_dendrites` function is defined by the time a `DendriticLayerBase` subclass is instantiated. Otherwise, it will only through an error much later on when `forward` is called.

nupic/research/frameworks/dendrites/dendritic_layers.py
```diff
@@ -143,6 +148,28 @@ def rezero_weights(self):
         if self.segments is not None:  # only none at beginning of init
             self.segments.rezero_weights()
 
+    def apply_dendrites(self, y, dendrite_activations):
```
mvacaporale, line 151: Please add an `@abc.abstractmethod` decorator here as well.

nupic/research/frameworks/dendrites/dendritic_layers.py
karangrewal: `DendriticLayerBase` needs to also be a subclass of `SparseWeights`, so it would actually be the following, correct?

```
class DendriticLayerBase(SparseWeights, metaclass=abc.ABCMeta):
```

mvacaporale: Ah true

karangrewal (approved):

PR #442, opened 2021-01-25
Experiments with subnetworks

lucasosouza:
* notebook with experiments on creating subnetworks with dendrites (RES-1912), and utils.py with supporting code
* a notebook tutorial for Vernon (RES-1931)

ps: utils.py, which is under a projects folder, uses the nomenclature adopted by scikit-learn of X and Y for data and target respectively. Hence the noqa on that file. 

ps2: since it is also just a projects file, I am not adding scikit-learn to nupic.research requirements.


projects/dendrites/subnetworks/utils.py
```diff
@@ -0,0 +1,383 @@
...
+        X, Y, test_size=0.2, random_state=random_state, stratify=Y)
+    return (X_train, X_test), (Y_train, Y_test)
+
+def apply_target_transform(Y, offset=0):
+    """Transform label for continual learning problems"""
+    return [label + offset for label in Y]
+
+
+def add_task_embedding(X, task_num=1, units=2):
+    """Adds a task embedding to the input"""
```
karangrewal, line 44: Can you add more detail in the docstring about what this function does?

projects/dendrites/subnetworks/utils.py
```diff
@@ -0,0 +1,383 @@
...
+        Xs, Ys = split_data(X, Y, random_state=random_state)
+        ready_datasets.append((Xs, Ys, dataset_name))
+
+    if len(ready_datasets) == 1:
+        return ready_datasets[0]
+    return ready_datasets
+
+
+def load_datasets():
+    """Load Iris, Wine and Mixed datasets"""
```
karangrewal, line 78: Can you please describe exactly what is returned by this function so that others will know how to call it appropriately?

projects/dendrites/subnetworks/utils.py
```diff
@@ -0,0 +1,383 @@
...
+    return (irisX, irisY, "Iris"), (wineX, wineY, "Wine"), (mixedX, mixedY, "Mixed")
+
+
+# ------ Training functions
+def evaluate(model, data, target, dataset_name=""):
+    pred = torch.argmax(model(torch.FloatTensor(data)), dim=1).numpy()
+    print(f"{dataset_name} Acc: {accuracy_score(target, pred):.4f}")
+
+def train(model, data, target, dataset_name, batch_size=1,
+          epochs=10, lr=1e-2, weight_decay=0, verbose=True):
```
karangrewal, line 115: how is this training function different from the one in [model utils](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/pytorch/model_utils.py#L37-L183), and why can't we re-use that?

projects/dendrites/subnetworks/utils.py
```diff
@@ -0,0 +1,383 @@
+# flake8: noqa
```
karangrewal, line 1: Please include the software license at the top of this file.

karangrewal (changes requested):
I requested some more documentation in a few areas, just to help future users who want to call your util functions.

Also, you defined `StandardMLP` and `DendriticMLP` which copy a lot of code, and the only difference between these and the networks defined in frameworks is additional batch normalization layers (as far as I could tell). I'm not sure if it's better to redefine these architectures here or in frameworks, so I will let you decide.

projects/dendrites/subnetworks/utils.py
lucasosouza: Not really different, but this is a small self-contained experiment. I needed the flexibility to modify the function as needed.

projects/dendrites/subnetworks/utils.py
lucasosouza: Good catch

projects/dendrites/subnetworks/utils.py
lucasosouza: Done

projects/dendrites/subnetworks/utils.py
lucasosouza: Done

lucasosouza:
Thanks @karangrewal , really good points. I incorporated your comments and replied above. I also fixed all styles issues in the document except for the variable names, as mentioned.

karangrewal (approved):

PR #448, opened 2021-02-05
Dendrites metric hook and mixin, and minor supporting changes

mvacaporale:
Major changes
- We now have full ability to create a variety of dendrite metrics.
- There's `ApplyDendritesHook` that collects `dendrite_activations` and the `winning_mask` of segments. This hook is then accessed at the end of each epoch for plotting relevant metrics.
- Plotting is done through the `PlotDendriteMetrics` mixin. The user passes any number of plot-functions into the config, each one taking arguments `dendrite_activations`, `winning_mask`, and `targets`.  These can also be configured to use fewer samples (helpful in some cases) and to plot every so many epochs (so that training isn't slowed down too much).
- An example config is included with this PR: `metacl_with_dendrite_stats` 

Minor changes
- Ensure hooks are copied over in `clone_model` for MetaCL experiments. 
- This change allows the metric mixins to work equally well on SurpervisedExperiments and MetaCL experiments. Thus there's a previously introduced mixin that has been removed.
- A mask utility for converting indices (from torch.max) to a mask.
- As much as possible, any new functionality also includes new tests to exercise it.
- There's a new base class `ApplyDendritesBase` to help distinguish the apply-dendrites modules. These are the ones we mostly care about for tracking dendrite metrics.
 
Warning: Using the `PlotDendriteMetrics` mixin with Ray is tricky. Namely, Ray throws an error when it inevitably tries to deepcopy a plot, which can't be done. After trying out a few options, the best I settled on is to simply wrap the plotting function and transform the plot into something more agreeable with Ray. See `prep_plot_for_wandb` in `ray_wandb.py` (an example is included in the aforementioned config). This function transforms a plot to a `wandb.Image`, which does work with Ray logging. In the future, we can add others such as `prep_plot_for_tensorboard` as needed.


mvacaporale:
I just added one last commit that logs the raw data whenever a dendrite metric is made. This way, we can always go back after training, adjust the plots, or even plot an entirely new metric.

nupic/research/frameworks/dendrites/apply_dendrites_hook.py
```diff
@@ -0,0 +1,84 @@
...
+        :param y: dendrite_output named tuple (with values and indices)
+        """
+
+        if not self._tracking:
+            return
+
+        dendrite_activations = x[1]
+        winning_mask = indices_to_mask(y.indices, shape=x[1].shape, dim=2)
+
+        # The `self` tensors were initialized on the cpu which could differ from the
```
lucasosouza, line 67: Why not initialize the self tensors directly on GPU using the device argument?

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
```diff
@@ -0,0 +1,247 @@
...
+                 wrapping the plot as done in `prep_plot_for_wandb`_.
+
+    .. _ApplyDendritesBase: nupic/reasearch/frameworks/dendrites/modules
+    .. _prep_plot_for_wandb: nupic/reasearch/frameworks/wandb/ray_wandb
+
+
+    :param config:
+        - plot_dendrite_metrics_args:
+
+            # The following arguments specify which modules of type `ApplyDendritesBase`
```
lucasosouza, line 68: The hashes can be removed here.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
```diff
@@ -0,0 +1,247 @@
...
+            # to track and create plots. The same modules are used for each 'plot_func'.
+            - include_modules: (optional) a list of module types to track
+            - include_names: (optional) a list of module names to track e.g.
+                             "features.stem"
+            - include_patterns: (optional) a list of regex patterns to compare to the
+                                names; for instance, all feature parameters in ResNet
+                                can be included through "features.*"
+
+            <plot_name>: This can be any string and maps to a dictionary of arguments
```
lucasosouza, line 77: I am confused with this docstring. Why is plot_name between brackets, and the arguments that follow are indented? Does this mean config is a dict, and a plot_name is a key within config that is also a dict and the arguments that follows are its keys?

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
```diff
@@ -0,0 +1,247 @@
...
+                targets = self.targets[:self.max_samples_to_track].cpu().numpy()
+                activations = activations[:self.max_samples_to_track].cpu().numpy()
+                winners = winners[:self.max_samples_to_track].cpu().numpy()
+                results.update({f"targets/{name}": targets})
+                results.update({f"dendrite_activations/{name}": activations})
+                results.update({f"winning_mask/{name}": winners})
+
+        return results
+
+    def error_loss(self, output, target, reduction="mean"):
```
lucasosouza, line 230: It feels weird to have the targets being collected during the loss function calculation, really going against the SRP. 

I understand you arrived at this solution because of the mix inheritance/functional approach currently in Vernon. If train_model was a class method instead of an outside function it would be easier to insert entry points that could be exploited by mixins.

But considering we will keep the structure as it is, one option would be to add an additional callback to train_model that could capture the target.

projects/meta_cl/experiments/metacl_track_stats.py
```diff
@@ -53,7 +65,42 @@ class TrackStatsMetaCLExperiment(mixins.TrackRepresentationSparsityMetaCL,
 )
 
 
+def plot_entropy_distribution_(_, winning_mask, targets):
```
lucasosouza, line 68: This pattern of using _ for unused arguments looks strange when used in the function definition. Is this a common thing used elsewhere? Maybe we should name the argument even if not using it in the code below.

lucasosouza (approved):
Left some comments, please feel free to disregard them if you don't think they are useful. The most relevant one is about capturing the targets in the loss function.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
```diff
@@ -0,0 +1,247 @@
...
+                                names; for instance, all feature parameters in ResNet
+                                can be included through "features.*"
+
+            <plot_name>: This can be any string and maps to a dictionary of arguments
+                         for each desired plot. The resulting plot will be logged under
+                         "<plot_name>/<module_name>" in the results dictionary.
+
+                - plot_func: the function called for plotting; must take three
+                             arguments: 'dendrite_activations', 'winning_mask', and
+                             'targets' (see above)
```
karangrewal, line 83: Recall that the plotting functions [here](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/dendrites/metrics/plotting.py) don't all need all 3 arguments. E.g., the function `plot_percent_active_dendrites` doesn't take `dendrite_activations`, so this seems inconsistent with what the docstring is saying.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
```diff
@@ -0,0 +1,247 @@
...
+                    continue
+
+                # Only use up the the max number of samples for plotting.
+                targets = self.targets[:max_samples_to_plot]
+                activations = activations[:max_samples_to_plot]
+                winners = winners[:max_samples_to_plot]
+
+                # Call and log the results of the plot function.
+                # Here, "{name}" is the name of the module.
+                visual = plot_func(activations, winners, targets, **plot_args)
```
karangrewal, line 215: This will cause an error if `plot_func` doesn't accept all 3 of 'activations', 'winners', 'targets' as inputs - which is the case for some plotting functions.

nupic/research/frameworks/dendrites/apply_dendrites_hook.py
mvacaporale: The hooks can be used just the same on cpu. However, it has no way of knowing which is the case when it's instantiated. Or at least it's best not to make assumptions.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
mvacaporale: I can see how this can be confusing. I'll add a docsting.
Here is an example:
    ```
    config=dict(
        plot_dendrite_metrics_args=dict(
            include_modules=[DendriticGate1d],
            mean_selected=dict(
                plot_func=mean_selected_activations,
            )
        )
    )
    ```

projects/meta_cl/experiments/metacl_track_stats.py
mvacaporale: Fair, I'll adjust it.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
mvacaporale: This is definitely the weakest part of the PR. It's a strict assumption, but not unreasonable, considering `error_loss` is explicitly passed to `train_model`. We took the path of least resistance. I'm tempted to keep it as is because it's been working just fine, and there are no major code changes needed to make it work. (If we edit `train_model`, then we also have to edit\check every call to `train_model`.)

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
mvacaporale: It supports the most general case which takes three arguments. My thinking was that we can adjust the signature of the functions as needed. (see below)

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
mvacaporale: In the example config, you'll notice it defines
```
def _plot_entropy_distribution(_dendrite_activations, winning_mask, targets):
    """
    This adjusts the function's signature to align with the mixin.
    """
    return plot_entropy_distribution(winning_mask, targets)
```

So whatever functions we define, we'll adjust the signature in a similar way and just take a dummy argument. I thought this was the most flexible way forward.

nupic/research/frameworks/vernon/mixins/dendrite_metrics.py
mvacaporale: The mixin supports the more general case that takes all three. I figure we can adjust the signatures as needed (see below).

karangrewal (approved):

PR #449, opened 2021-02-08
Updates to transformers

lucasosouza:
Includes run file and similar folder structure as in other nupic.research projects. To run, call `python run.py -e <experiment_name`. See README for detailed instructions on how to run it, including in distributed settings. 
* run.py and run_args.py: adapted from [examples](https://github.com/huggingface/transformers/blob/master/examples/language-modeling/run_mlm.py) in hugging face. Main modifications include:
1. accepting experiment dict as argument, using similar style as other nupic.research projects.
2. concatenating multiple datasets for language modeling, e.g. wikipedia and bookcorpus
3. custom auto saving and loading preprocessed datasets. This is redundant to HF smart caching, which is not working as expected (in smart caching the same datasets and flow of preprocessing should generate the same fingerprint - however the same script running twice in different instances is generating different fingerprints).
* run.sh: script to run in multiple nodes using torch.distributed.launch. Requires that instances are initialized using ray cluster tools. 
* scripts/: reorganized, added in bert_from_scratch.py option to run with ray (only tested single instance version). Default way to run now is using run.py, but I am keeping the script as a learning/introduction material to hugging face and transformers. 
* experiments/: includes some base experiments for debugging and a 100k steps BERT replication experiment.  

Edit: added functionality to incorporate trainer callbacks, and created a sparse BERT using callbacks to transform the fully connected hidden and output layer of each encoder block into sparse layers. See experiments/sparse_bert and callbacks/sparsity

To follow:
- Modify run.py to also perform finetuning given the appropriate config.
- Accept customized Trainer classes that ~enhance default Trainer using mixins~ incorporate callbacks [DONE]
- Run 100K dense Bert, finetune and evaluate; add results to Readme
- Create a 100k sparse Bert (first version), pretrain, finetune and evaluate; add results to Readme


lucasosouza:
@lscheinkman I am having issues with local vs circleci isort. I added the libraries to known-third-party in isort config - is that the correct way of fixing this?

I don't want to add these libraries to requirements just yet, as their dependencies might introduce breaking changes in our default environment.

subutai (changes requested):
Any code that is imported, other than experiment files, should not be in projects. Should be in frameworks/transformers or somewhere else under nupic.research.frameworks

lucasosouza:
> Any code that is imported, other than experiment files, should not be in projects. Should be in frameworks/transformers or somewhere else under nupic.research.frameworks

What do you mean imported, do you mean imported from external libraries? There is no class or method imported from nupic.research.frameworks. Breaking down run.py into two files (run and run_args) is just for code organization.

If we move this into frameworks now we will have to update requirements to use these external libraries, and I don't think this is a good thing at this point. The large number of dependencies can introduce breaking changes in our current projects (specially our ongoing sparse resnet project).

subutai:
> 
> If we move this into frameworks now we will have to update requirements to use these external libraries, and I don't think this is a good thing at this point. The large number of dependencies can introduce breaking changes in our current projects (specially our ongoing sparse resnet project).

Good point. I had not thought about that.

subutai (approved):
One minor change. Just took a very quick look, given that we will probably change this a lot.

PR #461, opened 2021-02-16
Mixin & Hook for plotting hidden activations

karangrewal:
This introduces

- a mixin class `PlotHiddenActivations` used for plotting hidden activations of specified modules,
- a forward hook class `TrackHiddenActivationsHook` (which subclasses 'TrackStatsHookBase') for keeping track of the hidden activations computed during forward passes,
- plotting functions to visualize the activations by unit, by class; note that these plotting functions don't take 'dendrite_activations' and 'winning_mask' as input since they have nothing to do with dendrites.

The resulting visualizations looks like [this](https://drive.google.com/file/d/1kQqn6m-Gx8viF2rnHAWDfUhzl8nB6nog/view?usp=sharing), and the mixin and hook classes are designed to mimic the `PlotDendriteMetrics` and `ApplyDendritesHook` classes, respectively.

### DIFF
--- a/nupic/research/frameworks/dendrites/metrics/metrics.py
+++ b/nupic/research/frameworks/dendrites/metrics/metrics.py
@@ -160,8 +160,11 @@ def dendrite_activations_by_unit(dendrite_activations, winning_mask, targets):
 def hidden_activations_by_unit(activations, targets):
     """
     Returns a 2D torch tensor with shape (num_categories, num_units) where cell c, i
-    gives the mean value (post-sigmoid) of the selected dendrite activation for unit i
-    over all given examples from category c.
+    gives the mean value of hidden activations for unit i over all given examples from
+    category c.
+
+    NOTE: This function is not specific to dendrites, and can be used with modules that
+    both include and don't include dendrite segments.
 
     :param activations: 2D torch tensor with shape (batch_size, num_units) where entry
                         b, i gives the activation of unit i for example b
--- a/nupic/research/frameworks/dendrites/metrics/metrics.py
+++ b/nupic/research/frameworks/dendrites/metrics/metrics.py
@@ -50,10 +50,10 @@ def percent_active_dendrites(winning_mask, targets):
         percent_active = percent_active.to(device)
 
         for t in range(num_categories):
-            inds_t = torch.nonzero(1.0 * (targets == t)).flatten()
-            num_examples_t = len(inds_t)
+            inds_t = torch.nonzero((targets == t).float(), as_tuple=True)
+            num_examples_t = len(inds_t[0])
 
-            percent_active_t = winning_mask[inds_t, :, :].sum(dim=0, dtype=torch.float)
+            percent_active_t = winning_mask[inds_t].sum(dim=0, dtype=torch.float)
             percent_active_t = percent_active_t / num_examples_t
 
             percent_active_t = percent_active_t.unsqueeze(2)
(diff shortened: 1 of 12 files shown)


nupic/research/frameworks/dendrites/metrics/metrics.py
```diff
@@ -157,6 +157,44 @@ def dendrite_activations_by_unit(dendrite_activations, winning_mask, targets):
         return selected_activations
 
 
+def hidden_activations_by_unit(activations, targets):
+    """
+    Returns a 2D torch tensor with shape (num_categories, num_units) where cell c, i
+    gives the mean value (post-sigmoid) of the selected dendrite activation for unit i
```
mvacaporale, line 163: When you say, "post-sigmoid", does this mean the input to this function comes from the output of a sigmoid?

nupic/research/frameworks/pytorch/hooks/track_hidden_activations.py
```diff
@@ -0,0 +1,60 @@
...
+        :param y: output of module
+        """
+        if not self._tracking:
+            return
+
+        self._activations = self._activations.to(y.device)
+        self._activations = torch.cat((y, self._activations), dim=0)
+
+        # Keep only the last 'num_samples'
+        self._activations = self._activations[:self.num_samples, ...]
```
mvacaporale, line 60: Related to my "post-sigmoid" question, do you intend to track a `torch.nn.Sigmoid` module with this hook? I know it could work for most modules; however, the reason I ask is to make sure you're not intending to track an `ApplyDendrites` module (e.g. `DendriticGate1d`) as these return a namedtuple for which you'd have to access via `y.values`.

nupic/research/frameworks/vernon/mixins/__init__.py
```diff
@@ -19,10 +19,15 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+from .composite_loss import CompositeLoss
```
mvacaporale, line 22: Thanks for organizing these.

nupic/research/frameworks/vernon/mixins/hidden_activations.py
```diff
@@ -0,0 +1,163 @@
...
+    TrackHiddenActivationsHook,
+)
+from nupic.research.frameworks.pytorch.model_utils import filter_modules
+
+__all__ = [
+    "PlotHiddenActivations",
+]
+
+
+class PlotHiddenActivations(metaclass=abc.ABCMeta):
```
mvacaporale, line 40: I'm glad this pattern of hook management and plotting is useful. There's definitely some repeated logic here. We don't need to worry about this much now. If anything, it would be useful however to include a test for this mixin just in case if we try and refactor the code to reduce repeated logic.

mvacaporale: See `dendrite_metric_mixins_test.py` for an example test. You can even just add to that file if you'd like.

mvacaporale (approved):
Looks good overall. I'd urge to add a test to the new mixin before merging, however. If we keep using this paradigm, I'll probably refactor the code later on, and a test will greatly help with that.

nupic/research/frameworks/dendrites/metrics/metrics.py
karangrewal: I think including this term is erroneous, since I mostly copied the description from `dendrite_activations_by_unit`. This has nothing to do with sigmoids, and I'll update the docstring.

The input to this function `activations` is just the output of a forward module (intended to be a dendritic layer or fc layer) over a batch of inputs. The parameter description for `activations` is correct.

nupic/research/frameworks/pytorch/hooks/track_hidden_activations.py
karangrewal: I don't intend to track a `torch.nn.Sigmoid` module, but it would work for that. In any case, I don't plan to use this hook for dendrites at all; it should only be used where the output is 2D, with shape (batch_size, num_features).

nupic/research/frameworks/vernon/mixins/hidden_activations.py
karangrewal: I'll add some test cases before merging.

karangrewal: Test cases have been added.

nupic/research/frameworks/dendrites/metrics/metrics.py
karangrewal: Updated.

PR #463, opened 2021-02-19
Representation overlap mixin

karangrewal:
The new mixin `PlotRepresentationOverlap` helps visualize overlaps in hidden representations. [Here](https://drive.google.com/file/d/1GU1eg25XG_qUmiGVzmYbTwc9eMI2GGtY/view?usp=sharing) is visual that shows the resulting figures. The figures on the left are the intra- and inter-class distributions of pairwise overlaps, and the figure on the right is a matrix where cell i, j gives the mean representation overlap in all pairs (x, y) where x belongs to class i, and y to class j.

This mixin class is very similar to `PlotDendriteMetrics` and `PlotHiddenActivations`. This mixin uses the `TrackHiddenActivationsHook` since this suffices for computing overlaps, instead of defining its own hook.

Additional functions to compute these metrics and plot them are included in the dendrites' metrics folder. Unit test cases are included.

### DIFF
--- a/nupic/research/frameworks/dendrites/metrics/metrics.py
+++ b/nupic/research/frameworks/dendrites/metrics/metrics.py
@@ -354,7 +354,7 @@ def representation_overlap_matrix(activations, targets):
         return overlap_matrix
 
 
-def representation_overlap_distributions(activations, targets):
+def representation_overlap_values(activations, targets):
     """
     Returns (list of floats, list of floats) where the first list gives representation
     overlap values between all pairs of samples in different classes, and the second
--- a/nupic/research/frameworks/dendrites/metrics/metrics.py
+++ b/nupic/research/frameworks/dendrites/metrics/metrics.py
@@ -21,6 +21,7 @@
 
 import math
 
+import numpy as np
 import torch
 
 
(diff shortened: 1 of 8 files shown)


karangrewal:
Tagging @alexcuozzo-1 so he can look over the source code for this new mixin. The future mixin that tracks gradient overlaps between classes will likely have similar functionality.

nupic/research/frameworks/dendrites/metrics/metrics.py
```diff
@@ -313,3 +314,126 @@ def entropy(x):
     _max_entropy = math.log(num_segments)
 
     return _entropy, _max_entropy
+
+
+def repr_overlap_matrix(activations, targets):
```
mvacaporale, line 319: Typically, we try and avoid shorthands for naming to be explicit when we can be. For local variables and arguments, I think it's often more acceptable to use shorthands/abbreviations since the context should be clear enough. Of course, we have no hard and fast rules.

If `repr` had a standard ML interpretation (like `conv`), your original would be a good name. But given our tendency to use full names when possible, and the fact that `repr` is a builtin python function, I'd vote for calling it `representaion_overlap_matrix`.

nupic/research/frameworks/dendrites/metrics/metrics.py
```diff
@@ -313,3 +314,126 @@ def entropy(x):
...
+        for t1 in range(num_categories):
+            for t2 in range(1 + t1):
+
+                mean_overlap_val = np.mean(repr_overlaps[t1][t2])
+                overlap_matrix[t1, t2] = mean_overlap_val
+
+        return overlap_matrix
+
+
+def repr_overlap_distributions(activations, targets):
```
mvacaporale, line 357: You're not returning the distributions, but the values that are used to populate histogram bins. It may make sense to call it `representation_overlap_scores` or `representation_overlap_values`.

nupic/research/frameworks/dendrites/metrics/plotting.py
```diff
@@ -440,11 +441,100 @@ def plot_entropy_distribution(winning_mask, targets):
...
+
+    for fig_num, ol in enumerate((inter_class_ol, intra_class_ol)):
+
+        plt.figure(fig_num)
+        plt.hist(x=ol, bins=np.arange(0.0, 1.0, 0.02), color="tab:blue",
+                 edgecolor="k")
+
+        plt.xticks(np.arange(0.0, 1.0, 0.2))
+        plt.xlabel("Fraction of overlap")
+        plt.ylabel("Pairwise frequency")
```
mvacaporale, line 527: I don't believe the histogram is plotting the frequency of occurrence. It's plotting the total number of occurrences, correct?

nupic/research/frameworks/vernon/mixins/representation_overlap.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Log the names of the modules being tracked
+        tracked_names = pformat(list(named_modules.keys()))
+        self.logger.info(f"Tracking representation overlap of modules: {tracked_names}")
+
+        # The targets will be collected in `self.error_loss` in a 1:1 fashion
+        # to the tensors being collected by the hooks.
+        self.ro_targets = torch.tensor([]).long()
+
+    def process_ro_args(self, ro_args):
```
mvacaporale, line 105: One thing you can do, so you don't have to distinguish the name of `process_args` with the `ro`, is to define `process_ro_args` as a stand-alone function in the same file. Which you can do since it doesn't require `self` anywhere in the function. This way, you won't have to worry about any conflicts with inheritance. Tbh, I should have done this with the mixins I wrote too.

nupic/research/frameworks/vernon/mixins/representation_overlap.py
```diff
@@ -0,0 +1,177 @@
...
+        This computes the loss and then saves the targets computed on this loss. This
+        mixin assumes these targets correspond, in a 1:1 fashion, to the samples seen
+        in the forward pass.
+        """
+        loss = super().error_loss(output, target, reduction=reduction)
+        if self.ro_hook.tracking:
+
+            # Targets were initialized on the cpu which could differ from the
+            # targets collected during the forward pass.
+            self.ro_targets = self.ro_targets.to(target.device)
```
mvacaporale, line 171: Sidenote: I understand why you need the name `ro_targets`, as to not conflict with the other `self.targets` definition.  In your last PR, if you recall, I mentioned that it could be good to generalize the functionality so we didn't need to keep repeating mixins over and over. That would also help with this naming conflict.

If you want, it could be a good exercise for you try and make such a general mixin. It basically needs to pair a hook, the modules that hook will track, and the functions it will plot for each hook. Ideally, the mixin could take a list of hooks and plot for each hook. 

No pressure though. We can do without. I'd prefer you optimize for your advancing the research questions you're going down. However, if you keep finding yourself writing similar mixins, then we can discuss how this could be done; we're not too far from it already.

mvacaporale (approved):

mvacaporale:
Looks good, just a few minor comments.

nupic/research/frameworks/dendrites/metrics/metrics.py
karangrewal: Fixed.

nupic/research/frameworks/dendrites/metrics/metrics.py
karangrewal: Good point. Done.

nupic/research/frameworks/dendrites/metrics/plotting.py
karangrewal: Fixed.

nupic/research/frameworks/vernon/mixins/representation_overlap.py
karangrewal: I agree; it's now a separate function.

nupic/research/frameworks/vernon/mixins/representation_overlap.py
karangrewal: I agree this would be useful, but I'll leave `ro_targets` as is for now, to keep it simple.

PR #467, opened 2021-02-24
Various GSC related updates

subutai:


projects/gsc/experiments/base.py
```diff
@@ -128,11 +131,12 @@
     ),
 
     # Loss function. See "torch.nn.functional"
-    loss_function=torch.nn.functional.cross_entropy,
+    loss_function=torch.nn.functional.nll_loss,
```
mvacaporale, line 134: This was the main thing to change given our previous discussion. There's also `batch_norm_weight_decay` which now has no effect given [this](https://github.com/numenta/nupic.research/pull/460) PR (To control the batch norm weight decay, you'd need the `ConfigureOptimizerParamGroups` which isn't currently being used by the GSC experiment).

This is all to say, `batch_norm_weight_decay` should be removed from this config as it has no effect on training. As well, is the desired behavior per our discussion (weight decay applied to all params including batchnorm).

mvacaporale: Minor point: You can increase the size of `val_batch_size`. Right now it's 16, but a larger size could speed up inference.

projects/gsc/experiments/base.py
```diff
@@ -177,11 +181,53 @@
     batch_sizes=[4, 16],  # 4 for the first epoch and 16 for the remaining
     val_batch_size=1000,
     batches_in_epoch=sys.maxsize,
+)
```
mvacaporale, line 184: Just a side note: `DEFAULT_BASE` should replicate the 97% results. I'm not sure about `DEFAULT_SPARSE_CNN`. In our discussion, we made sure `DEFAULT_BASE` replicated the best result from the `whydense` project. Could be helpful to make sure this is clear in this experiment file if not already.

mvacaporale (approved):

PR #470, opened 2021-02-26
Sparse Transformers: new models, updated checkpoints, and new configs.

mvacaporale:
**Per the new models**
- There are two new models that sparsify either the whole encoder or just the non-attention layers.
- There's a class wrapper `register_bert_model` for Bert models that automatically creates a related `BertConfig`, `BertForMaskedLM` , and `BertForSequenceClassification` and then registers these classes with the Transformers library so they may be loaded by `AutoConfig`, `AutoModelForMaskedLM`, and `AutoModelForSequenceClassification`. More on this below.

**Per the checkpoints:**
- I modified our previously made checkpoint in `bert-steps_100k-sparsity_0.8` to use one of those new models so we can load it more easily. This model is in `bert-steps_100k-sparsity_0.8_updated` 
- I exported this checkpoint to `/mnt/efs/results/pretrained-models/transformers-local/static_sparse_non_attention_bert_100k` 

**Per the configs:**
- The config `finetuning_sparse_bert_100k_glue` finetunes that model and is in the updated README under `sparse_v1_100k` 
- Currently, I'm running `static_sparse_encoder_bert_100k` which sparsifies all layers, including attention. I'll update those results soon.

**Example with `register_bert_model`:**
```
@register_bert_model
class SparseBertModel(BertModel):

    @dataclass
    class ConfigKWargs:
        # Keyword arguments to configure sparsity.
        sparsity: float = 0.9

    # Define __init__, ect.
    ...
```
This will automatically create new classes called `SparseBertConfig`, `SparseBertForMaskedLM`, and `SparseBertForSequenceClassification`. Notice that the naming is automatic and is a function of the name of your original class. For instance, if you define `DynamicSparseBertModel`, you'd get a class named `DynamicSparseBertConfig` and so on.

As soon as you define the class, it's ready to autoload. For instance, you could do
```
config = AutoConfig.for_model(model_type="sparse_bert", sparsity=0.5)
model = AutoModelForMaskedLM.from_config(model)

type(model)
>>> SparseBertModelForMaskedLM
```

Notice how the `model_type` of `"sparse_bert"` has also been automatically formatted. In the other example, you'd use `model_type="dynamic_sparse_bert"`. As well, the config is already equipped to accept the argument `sparsity` which can be accessed by your model. Thus, you can run
```
config.sparsity
>>> 0.5
```
This comes from the `ConfigKWargs` defined above. You can add whatever arguments you want to that dataclass. With this, we can modify our experiment configs and configure our models as desired.


projects/transformers/callbacks/sparsity.py
```diff
@@ -19,10 +19,20 @@
...
 
+logger = logging.get_logger(__name__)
+
+__all__ = [
+    "RezeroWeightsCallback",
+    "SparsifyFCLayersCallback",
+]
+
 
 class SparsifyFCLayersCallback(TrainerCallback):
```
lucasosouza, line 37: Should we keep this? I am ok with removing it if we have a better way not, that can avoid confusion.

projects/transformers/callbacks/sparsity.py
```diff
@@ -49,3 +59,40 @@ def on_init_end(self, args, state, control, model, **kwargs):
...
+
+
+class RezeroWeightsCallback(TrainerCallback):
+
+    def on_init_end(self, args, state, control, model, **kwargs):
+        """Log sparsity of the model and the sparsity of just the encoder."""
+
+        model.apply(rezero_weights)
+
+        num_tol, num_nonzero = count_nonzero_params(model)
```
lucasosouza, line 71: Is num_tol a typo? It feels like you wanted to write num_total

projects/transformers/experiments/sparse_bert.py
```diff
@@ -48,8 +48,34 @@
 )
 
 
+static_sparse_encoder_bert_100k = deepcopy(bert_100k)
+static_sparse_encoder_bert_100k.update(
+    # Model Arguments
+    overwrite_output_dir=False,
+    model_type="static_sparse_encoder_bert",
+    config_kwargs=dict(
```
lucasosouza, line 56: Is the plan to use the same model, but keep evolving so config takes more parameters, like to specify which layers to sparsify and which method to use for sparsification ? (eg random uniform by layer, random uniform by head, or some other form of sparsity initialization).

projects/transformers/experiments/sparse_bert.py
```diff
@@ -48,8 +48,34 @@
...
+        sparsity=0.8,
+    ),
+    trainer_callbacks=[RezeroWeightsCallback()]
+)
+
+# Sparse Bert of only two layers and one attention head.
+mini_sparse_bert_debug = deepcopy(debug_bert)
+mini_sparse_bert_debug.update(
+    model_type="static_sparse_encoder_bert",
+    config_kwargs=dict(
```
lucasosouza, line 66: Not changing the hidden size of intermediate and output layers? The ratio of number of weights in attention layers to number of weights in non-attention layers will be significantly unbalanced if you don't change those as well (see the calculations we did yesterday).

projects/transformers/models/register_bert_model.py
```diff
@@ -0,0 +1,279 @@
...
+)
+from transformers.models.bert.modeling_bert import BertOnlyMLMHead
+from transformers.utils import logging
+
+# This is the dict of the `models` module. Anything imported from `models`
+# must be contained in this dict. Hence, it will be updated to include
+# the new modules made below so they may be imported.
+from . import __dict__ as __models_dict__
+
+logger = logging.get_logger(__name__)
```
lucasosouza, line 45: Did you meant to redefine logger? My understanding is these logs will not show when calling from run.py if you redefine it here. Let me know if I am wrong about this.

projects/transformers/run_utils.py
```diff
@@ -610,6 +610,7 @@ def init_tokenizer(model_args):
         use_fast=model_args.use_fast_tokenizer,
         revision=model_args.model_revision,
         use_auth_token=True if model_args.use_auth_token else None,
+        config=config,
```
lucasosouza, line 613: This is interesting. How is the tokenizer affected by the changes made to sparsify the model? Shouldn't we use the same tokenizer used for both cases?

If we change the tokenizer, the caching tokenized dataset process will silently fail, since today is only based on dataset+dataset_config and assumes there will be no changes to the tokenizer (for example if we change the number of tokens in the model embedding layer). Keep in mind tokenization is very costly and it will take almost 2 days to re-tokenize wikipedia, hence the forced loading from cache style. If we move to larger datasets that is going to be even worse. 

Not requiring immediate changes, but just want to raise awareness about this issue since there is potential to introduce bugs if we are passing this config to the tokenizer. We should think about changing the behavior in run.py and accept two different model arguments, one to the tokenizer which will be fixed to `bert-based-case` unless specifically defined otherwise, and the other to the model.

projects/transformers/tests/sparse_bert_model_test.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class SparseBertModelTest(unittest.TestCase):
+    """
+    Test the ability to load and save StaticSparseEncoderBert models and associated
+    StaticSparseEncoderBertConfig.
+    """
+
+    def setUp(self):
+        self.config = CONFIG_MAPPING["static_sparse_encoder_bert"](
```
lucasosouza, line 52: Same comment as above, this network is actually quite big and disproportional in terms of number of weights to attention and to non-attention layers. It has zero impact on test except for speed, but if we change in experiments we might as well reflect those changes here.

projects/transformers/export_model.py
```diff
@@ -31,13 +31,16 @@
 
 from transformers import AutoModelForMaskedLM
 
+# Import models. This will update Transformer's model mappings so that custom models can
+# be loaded via AutoModelForMaskedLM.
```
lucasosouza, line 35: Not sure I got this magic. Can you please explain it how this works in our next call?

lucasosouza (approved):
Nice job. It seems to run, but I have some questions below. I am specifically concerned with passing arguments to the tokenizer.

projects/transformers/callbacks/sparsity.py
mvacaporale: I can remove it, but we'll need to update the `sparse_bert_100k` to use one of the new models. The architecture will be the same, just different functionality will load it.

projects/transformers/callbacks/sparsity.py
mvacaporale: Tried to go for a shorthand but messed up there.

projects/transformers/experiments/sparse_bert.py
mvacaporale: We could do that. I think that will work best for minor tweaks (e.g. adjusting the sparse init). We can also make new models as needed. For instance, we can probably make a `VDropBertModel` or something like that at some point.

projects/transformers/experiments/sparse_bert.py
mvacaporale: This was just a debug model to make sure things can run. Do you think adjusting those layers is needed?

projects/transformers/callbacks/sparsity.py
lucasosouza: Your call if it is worth the trouble or not. If we don't do it we can at least add a note to the docstring saying it is deprecated.

projects/transformers/experiments/sparse_bert.py
lucasosouza: If is a debug thing, won't matter (except again for speed). But it won't hurt to add two extra parameters and can avoid misleading someone new to the project reading the code and looking for an example of a mini sparse bert.

projects/transformers/models/register_bert_model.py
mvacaporale: I copied this setup from the transformers repo. That's how they do it in most (if not all) files I've seen. I just did a quick test. The logs show up just fine when running `run.py`.

projects/transformers/export_model.py
```diff
@@ -31,13 +31,16 @@
...
+import models # noqa
```
lucasosouza, line 36: I think the rule you are looking for here is F401, so # noqa: F401.

For completeness, a different way of doing this without the noqa: https://stackoverflow.com/questions/59167405/flake8-ignore-only-f401-rule-in-entire-file

projects/transformers/models/register_bert_model.py
lucasosouza: Does it picks up up the changes in config applied in run.py?

lucasosouza: Referring to 
```
        logging.basicConfig(
            format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s",
            datefmt="%m/%d/%Y %H:%M:%S",
            handlers=[logging.StreamHandler(sys.stdout)],
            level=(logging.INFO if is_main_process(training_args.local_rank)
                   else logging.WARN)
        )
```

projects/transformers/experiments/sparse_bert.py
mvacaporale: Fair enough. Still a bit shaky in my understanding in how the number of attention heads relates to the hidden size. Can you explain again?

projects/transformers/models/register_bert_model.py
lucasosouza: Reason why I'm asking is I had issues hence changed. I don't clearly remember what was the bug, I think it iwas in the distributed setting. See commit: https://github.com/numenta/nupic.research/commit/d5a944de1cce99881db9f41e0a177edf834294ca.

If you are confident this works fine, we should change it in every file, instead of having half of the files one way, and half of the files the other way.

projects/transformers/export_model.py
mvacaporale: Sure thing. The basic idea is that our custom sparse models get registered in the appropriate dictionaries so that functions like `AutoModelForMaskedLM` can use them. This happens as soon as they are defined thanks to `register_bert_model`. And they get defined as soon as you import the `models` module.

projects/transformers/experiments/sparse_bert.py
lucasosouza: If you change num_attention_heads and look the num of params, they are the same. I think that is just defining in how many heads the weights are divided by. Would have to go into the model code to confirm this hypothesis. 

So you have to change the hidden_size to actually change the number of params. Then if you change the hidden size without changing the intermediate size, you will have a disproportionate amount of weights in the linear transformations post attention compared to attention. 

I didn't spend time looking at the model, I am assuming you would that.

projects/transformers/export_model.py
mvacaporale: Added it.

projects/transformers/tests/sparse_bert_model_test.py
mvacaporale: I changed it, mostly for speed of the test.

projects/transformers/models/register_bert_model.py
mvacaporale: 1. It does not use the format `format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s"`
2. I don't know if it works in the distributed setting.
3. Could be could to change it (not sure which is better yet). I can add a TODO if that's alright.

lucasosouza: Can't we just stick it to the way it is right now instead, I don't see why change it if we don't know if the new way works

projects/transformers/run_utils.py
mvacaporale: Good point. I'm not sure why exactly I added it. Possibly just because the Tokenizer can accept the config. I just ran it without this line, and no errors are introduced. So I think I'll remove it just to be safe.

projects/transformers/models/register_bert_model.py
mvacaporale: Sure. It may not make a difference, but I do see the benefit in reducing the potential for errors.

projects/transformers/experiments/sparse_bert.py
mvacaporale: I see. That helps. Particularly, "If you change num_attention_heads and look the num of params, they are the same". I double-checked, you're right about this. I'll update the mini model.

projects/transformers/export_model.py
lucasosouza: If it is changing a constant in the global namespace, does that mean it is dependent on the order it is imported? If we change the import order, does it still work as expected?

lucasosouza: @mvacaporale just making sure you saw this question, that is still not clear to me.

PR #475, opened 2021-03-10
Bertitos and other Transformer Updates

mvacaporale:
There are new reduced sized Bert baselines (see review of results [here](https://wandb.ai/numenta/huggingface/reports/Bertitos-Baselines--Vmlldzo1MTk2MzY)).

Related: included is a notebook to create and tokenized a small dataset. It includes 1% Wikipedia and 8% of Book Corpus
for a total of 1,132,942 text samples once tokenized with a `max_seq_length` of 128. This is roughly 8.7 gigabytes of tokenized data total. To use it, we set
```
    max_seq_length=128,
    dataset_name="wikipedia_plus_bookcorpus",
    dataset_config_name=None,
``` 

Updates include:
- Dataset names are now hashed with `max_seq_length` in the name; before they just included the dataset names and dataset-config names
- There's a new `CustomWandbCallback`. This is used to update the callback mapped from `wandb` in `transformers.integrations.INTEGRATION_TO_CALLBACK`. Now, wandb will automatically log perplexity (we can include other enhancements later as well). Note, I looked into initializing wandb earlier in training so we can view fuller logs in wandb; however, it seemed a bit involved, so I left it for later.
- New finetuning configs (I forgot to include these in the last PR)


projects/transformers/README.md
```diff
@@ -4,17 +4,17 @@
...
-| sparse_v1_100k |          72.67 |          67.91 |  22.43 | 77.25/77.96 | 85.37/78.82 |  84.65 | 88.23/84.32 | 58.32 |  87.27 | 82.79/82.65 |  29.84 | 10.461 |
-| sparse_v2_100k |          68.41 |          67.06 |  15.96 | 72.25/73.11 | 80.50/71.35 |  81.23 | 85.78/81.01 | 58.98 |  82.64 | 76.53/76.40 |  56.25 | 16.696 |
-| sparse_v3_100k |          68.25 |          66.92 |  16.35 | 72.00/72.32 | 80.42/70.83 |  80.31 | 85.04/80.49 | 58.98 |  82.52 | 76.96/77.60 |  56.25 | 22.81 |
+|            | average_bert     | average_glue      | cola           | mnli                         | mrpc        | qnli     | qqp         | rte      | sst2     | stsb                  | wnli     | perplexity | eval loss |
+|:-----------|:-----------------|:------------------|:---------------|:-----------------------------|:------------|:---------|:------------|:---------|:---------|:----------------------|:---------|:-----------|:----------|
+|            | Average w/o wnli | Average all tasks | Matthew's corr | Matched acc./Mismatched acc. | F1/Accuracy | Accuracy | Accuracy/F1 | Accuracy | Accuracy | Person/Spearman corr. | Accuracy |            | log(perplexoty) |
```
lucasosouza, line 9: typo: log(perplexity)

projects/transformers/run.py
```diff
@@ -49,6 +49,7 @@
 )
 from transformers.trainer_utils import get_last_checkpoint, is_main_process
 
+import integrations  # noqa
```
lucasosouza, line 52: Would be better to specify which rule you need to skip, if possible.

projects/transformers/integrations.py
```diff
@@ -0,0 +1,54 @@
...
+from transformers.integrations import INTEGRATION_TO_CALLBACK, WandbCallback
+
+
+"""
+This file serves to extend the functionalities of automated integrations such as those
+for wandb.
+"""
+
+
+class CustomWandbCallback(WandbCallback):
```
lucasosouza, line 34: Given this callback replaces the original one in wandb, for any metrics we want to track we should modify this class?

lucasosouza (approved):
Looks good!

projects/transformers/experiments/bertitos.py
```diff
@@ -0,0 +1,222 @@
...
+# |---------------------------------|
+# | model           | training time |
+# |-----------------|:-------------:|
+# | small_bert_100k | ~8 hrs        |
+# | mini_bert_100k  | ~6 hrs        |
+# | tiny_bert_100k  | ~2 hrs 20 min |
+# |---------------------------------|
+# (reported using a one p3.2xlarge)
+#
+
```
lucasosouza, line 70: It would be useful for quick reference to add one line per method with a summary of each bertito - num attention heads, hidden layers, hidden size and intermediate size. That could be part of the second table for example.

projects/transformers/integrations.py
mvacaporale: Potentially. Some metrics may be better suited for a different callback. I think it will depend. I see the `CustomWandbCallback` being suitable for a metric that applies to all experiments. On the other hand, something that tracks sparsity can log to wandb directly by calling `wandb.log`.

PR #486, opened 2021-03-29
Address issues with wandb early init as well as transformers dependencies 

mvacaporale:
This fixes a bug where run.py tries to initialize wandb even if it's not installed. This also updates the wandb callback to work with the latest transformers version (`4.5.0.dev0`).

The versions for transformers, datasets, wandb, and torch are now specified in a  `requirements.txt` file so they may all be install via `pip install -r requirements.txt`.

Feedback is welcome on the versions. The goal here is to make sure they are detailed enough to avoid bugs (like the wandb bug addressed in this PR) while still making new transformers and torch features readily available.

### DIFF
--- a/projects/transformers/requirements.txt
+++ b/projects/transformers/requirements.txt
@@ -20,6 +20,9 @@
 #
 # ------------------------------------------------------------------------------
 # Run pip install -r requirements.txt to reproduce the currently tested environment
-# Last update: 2/17/2021
--e git+https://github.com/huggingface/transformers@e94d63f6cbf5efe288e41d9840a96a5857090617#egg=transformers
--e git+https://github.com/huggingface/datasets@58555b2734b8a18d70b5a2418129f26fedd96a39#egg=datasets
+# Last update: 3/29/2021
+
+git+git:https://github.com/huggingface/transformers@master#egg=transformers
+datasets>=1.5.0
+torch>=1.8.1
+wandb>=0.10.23
--- a/projects/transformers/requirements.txt
+++ b/projects/transformers/requirements.txt
@@ -26,3 +26,5 @@ git+git:https://github.com/huggingface/transformers@master#egg=transformers
 datasets>=1.5.0
 torch>=1.8.1
 wandb>=0.10.23
+ray>=1.2.0
+pickle5>=0.0.11
--- a/projects/transformers/run.py
+++ b/projects/transformers/run.py
@@ -48,6 +48,7 @@ from transformers import (
     default_data_collator,
     set_seed,
 )
+from transformers.integrations import is_wandb_available
 from transformers.trainer_utils import get_last_checkpoint, is_main_process
 
 from experiments import CONFIGS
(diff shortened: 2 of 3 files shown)


projects/transformers/requirements.txt
```diff
@@ -20,6 +20,9 @@
 #
 # ------------------------------------------------------------------------------
 # Run pip install -r requirements.txt to reproduce the currently tested environment
-# Last update: 2/17/2021
--e git+https://github.com/huggingface/transformers@e94d63f6cbf5efe288e41d9840a96a5857090617#egg=transformers
--e git+https://github.com/huggingface/datasets@58555b2734b8a18d70b5a2418129f26fedd96a39#egg=datasets
\ No newline at end of file
+# Last update: 3/29/2021
+
+git+git:https://github.com/huggingface/transformers@master#egg=transformers
```
lscheinkman, line 25: The PR description mentions `4.5.0.dev0` but this code will install the latest version from `master`. If the goal is stability then you should update the commit to `4.5.0.dev0` instead of plain `master` branch

projects/transformers/run.py
```diff
@@ -105,7 +106,7 @@ def main():
 
         # Initialize wandb now to include the logs that follow.
         # For now, only support early wandb logging when running one experiment.
-        if len(cmd_args.experiments) == 1:
+        if is_wandb_available() and len(cmd_args.experiments) == 1:
```
lucasosouza, line 109: This looks good, does it fix the WAND_DISABLED=true issue as well?

projects/transformers/requirements.txt
lucasosouza: I am fine either the way, but as reminder we can't fix transformers without fixing datasets and tokenizers - we had issues in the past with mismatch, and they don't guarantee the latest versions in pip are synced (they do guarantee the latest versions from source are synced)

projects/transformers/requirements.txt
```diff
@@ -20,6 +20,9 @@
...
+datasets>=1.5.0
+torch>=1.8.1
+wandb>=0.10.23
```
lucasosouza, line 28: Since you are on it, can we add ray latest and pickle5 latest to the list - they are required to run the hp search (https://github.com/numenta/nupic.research/pull/485)

mvacaporale: Done

projects/transformers/run.py
mvacaporale: Yes, it does. I was able to reproduce that bug and fix it by editing the wandb callback class.

lucasosouza: Isn't is_wandb_available() still gonna check for WANDB_DISABLED and return False, hence throwing the same error?

projects/transformers/requirements.txt
mvacaporale: We're balancing two motivations: hedging risk against errors caused by misaligned dependencies and having the latest features of the Huggingface repos. We've been prioritizing the latter for a faster pace of development (as we're actively waiting for planned upcoming features). I'll keep these dependencies as is, but we can restrict them more as needed down the line.

projects/transformers/run.py
mvacaporale: That error was thrown by an assert, whereas this is an if-statement. Thus, we won't get to that assert if this if-statement isn't passed first.

projects/transformers/requirements.txt
lscheinkman: Just keep in mind `master` is a moving target and the code installed will depend on the day you call `pip install`

lucasosouza (approved):

PR #493, opened 2021-04-06
Remove use of `trainer_extra_kwargs`

mvacaporale:
This is a refactoring PR that replaces `trainer_extra_kwargs`, originally passed through `model_args`, to `mixin_args` which is now passed through `training_args`. This has two benefits:
1. These args are now log-able to wandb (they will show up under the run's config).
2. We can run an hp search over mixin related arguments.

Before, 2. was not doable since the hp_space method was only for updating the trainers `self.args`. Now `mixin_args` are a subset of these and can be tuned as needed. As well, the code should be a bit cleaner and easier to use (hopefully).

### DIFF
--- a/projects/transformers/trainer_mixins/distillation.py
+++ b/projects/transformers/trainer_mixins/distillation.py
@@ -23,6 +23,7 @@ import logging
 
 import torch
 import torch.nn.functional as F
+from transformers import AutoModelForMaskedLM
 
 from nupic.research.frameworks.pytorch.model_utils import (
     filter_modules,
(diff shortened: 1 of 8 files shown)


projects/transformers/trainer_mixins/distillation.py
```diff
@@ -86,16 +79,34 @@ def __init__(
...
         # Validate teacher models
         assert (
-            teacher_models is not None
-        ), "When using KD mixin, teacher_models must be defined"
+            isinstance(teacher_names_or_paths, list) and len(teacher_names_or_paths) > 0
+        ), "When using KD mixin, teacher_models_name_or_path must be defined"
+
+        teacher_models = []
+        for model_name_or_path in teacher_names_or_paths:
+            teacher_model = AutoModelForMaskedLM.from_pretrained(
```
lucasosouza, line 101: This doesn't include model_kwargs. I see that most of the arguments are irrelevant in our case, but are we sure config is redundant and can be removed?

projects/transformers/trainer_mixins/distillation.py
```diff
@@ -41,22 +42,14 @@ class DistillationTrainerMixin:
...
-        kd_factor_init=1.0,
-        kd_factor_end=1.0,
-        kd_temperature_init=1.0,
-        kd_temperature_end=1.0,
-        **kwargs,
-    ):
+    def __init__(self, *args, **kwargs):
         """
-        Add following variables to Trainer
+        Add following variables under 'mixin_args' through the training arguments.
```
lucasosouza, line 47: Should be trainer_mixin. We might still need model_mixins (or maybe even mixins to the tokenizer class, for example)

lucasosouza:
Please see comments

projects/transformers/trainer_mixins/distillation.py
mvacaporale: I checked some of the args, and I just checked the others. Here's the rundown: 
* `config` will be automatically instantiated through the `from_pretrained` command
* `cache_dir` is still helpful and is passable through `teacher_model_cache_dir`
* `revision` could be helpful, but so far we haven't across a scenario where it is
* `use_auth_token`, this is used for private models saved to HF, which we don't have

So we should be good to go.

lucasosouza: I was just worried about config, but if it is automatically instantiated then that makes it redundant.

lucasosouza (approved):

projects/transformers/trainer_mixins/distillation.py
mvacaporale: Fair point, just changed it.

PR #497, opened 2021-04-09
Sparsity parameter bug fix and unit tests for SparseMLP. 

akashvelu:
Fixes sparsity parameter bug in add_sparse_linear_layer function, change variable names, and add tests for SparseMLP.


nupic/research/frameworks/pytorch/models/le_sparse_net.py
```diff
@@ -153,7 +153,11 @@ def add_sparse_linear_layer(
             )
         else:
             network.add_module(
-                "linear{}".format(suffix), SparseWeights(linear, weight_sparsity)
+                "linear{}".format(suffix),
```
subutai, line 156: I don't think this change is correct. Here, the parameter is defined as "Pct of weights that are allowed to be non-zero". You would need to use `sparsity = 1 - weight_sparsity`

subutai (changes requested):

nupic/research/frameworks/pytorch/models/le_sparse_net.py
akashvelu: Thanks for catching that, fixed.

subutai (approved):

lucasosouza (approved):

PR #499, opened 2021-04-11
Move DendriticMLP into frameworks/dendrites

subutai:
Moved all the weight initialization stuff to be inside the class since they access internal details and member properties.


nupic/research/frameworks/dendrites/modules/dendritic_mlp.py
```diff
@@ -0,0 +1,225 @@
...
+        output = self.kw1(output) if self.kw else output
+
+        output = self.dend2(output, context=context)
+        output = self.kw2(output) if self.kw else output
+
+        output = self.classifier(output)
+        return output
+
+    # ------ Weight initialization functions
+    def _init_sparse_weights(self, m, input_sparsity):
```
karangrewal, line 131: Why start the function names of `_init_sparse_weights` and `_init_sparse_dendrites` with underscore?

karangrewal: Also, I don't see any benefit to having these as methods instead of functions, since they never refer to `self`. We might as well keep them as stand-alone functions.

nupic/research/frameworks/dendrites/modules/dendritic_mlp.py
```diff
@@ -0,0 +1,225 @@
...
+            the unit is turned off for any other context - they contain negative weights
+            for all other weights.
+
+        :param context_vectors:
+        :param init: a string "overlapping" or "non_overlapping"
+        """
+        self._hardcode_dendritic_weights(self.dend1.segments, context_vectors, init)
+        self._hardcode_dendritic_weights(self.dend2.segments, context_vectors, init)
+
+    def _hardcode_dendritic_weights(self, dendrite_segments, context_vectors, init):
```
karangrewal, line 181: `_hardcode_dendritic_weights` doesn't use `self`, so once again it's unnecessary to have it as a method rather than a separate function. We should collapse hardcoding into a single method, or move this 2nd method to a stand-alone function.

karangrewal (changes requested):

nupic/research/frameworks/dendrites/modules/dendritic_mlp.py
subutai: The underscore is a python convention - it refers to the fact that it's an internal only method (external code shouldn't call this method). Since it isn't using self, I can make it a static method. It should be inside the class however because it is accessing details about the class that external callers shouldn't access.

nupic/research/frameworks/dendrites/modules/dendritic_mlp.py
subutai: Same answer as before. I'll make it into a static method. I don't think we can make it a single method because it's called twice in hardcode_dendritic_weights()

subutai:
BTW, I replicated the 50 task experiment and it gets the same answer as before: 91.5%

karangrewal (approved):

PR #501, opened 2021-04-14
permutedMNIST

karangrewal:
This PR mainly concerns the addition of the permutedMNIST dataset for continual learning. For memory efficiency, the MNIST images are not duplicated, and instead the different permutations are held in memory and applied to each image when being loaded.

Other notable changes:

1. In addition to `PermutedMNIST`, there's also a `ContextDependentPermutedMNIST` dataset class which augments the former by additionally returning a binary sparse context vector. This is useful in the case of dendritic networks when task id is provided and not inferred.
2. `compute_task_indices` in the CL experiment class has been slightly modified so that train/test loaders need not yield 2-item tuples (as was previously assumed), but now can yield any number of batch items as long as the target label is the last one.
3. `DendriticMLP` has been augmented with the `dendrite_sparsity` parameter, so users may specify the level of sparsity in dendritic weights. This is useful because in many experiments, forgetting is only avoided when dendrites are fully dense. This change to `DendriticMLP` has been propagated throughout nupic.research.
4. The new project file in the permutedMNIST project folder runs a dendritic MLP on permutedMNIST (with context provided) in an identical setup to the [XdG model](https://www.pnas.org/content/pnas/115/44/E10467.full.pdf).

### DIFF
--- a/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
+++ b/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
@@ -109,11 +109,11 @@ class ContextDependentPermutedMNIST(PermutedMNIST):
 
     def __getitem__(self, index):
         """
-        Returns an (image, context, target) triplet.
+        Returns an ((image, context), target) tuple.
         """
         img, target = super().__getitem__(index)
         task_id = self.get_task_id(index)
-        return img, self.contexts[task_id, :], target
+        return (img, self.contexts[task_id, :]), target
 
     def init_contexts(self, seed):
         percent_on = 0.05
--- a/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
+++ b/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
@@ -25,8 +25,6 @@ import torch
 from torchvision import transforms
 from torchvision.datasets import MNIST
 
-from nupic.research.frameworks.dendrites.routing import generate_context_vectors
-
 
 class PermutedMNIST(MNIST):
     """
@@ -38,11 +36,7 @@ class PermutedMNIST(MNIST):
     permutation) that a continual learner will try to learn in sequence.
     """
 
-    # Permutations should be the same whether this class is instantiated for training
-    # or testing
-    permutations = None
-
-    def __init__(self, num_tasks, train, root=".", target_transform=None,
+    def __init__(self, num_tasks, seed, train, root=".", target_transform=None,
                  download=False):
 
         data_transform = transforms.Compose([
@@ -54,15 +48,16 @@ class PermutedMNIST(MNIST):
 
         self.num_tasks = num_tasks
 
-        # Generate num_tasks random permutations - the first one is the identity
-        # permutation (i.e., regular MNIST), represented below as `None`
-        if PermutedMNIST.permutations is None:
-            PermutedMNIST.permutations = [
-                torch.randperm(784) for task_id in range(1, num_tasks)
-            ]
-            PermutedMNIST.permutations.insert(0, None)
+        # Use a generator object to manually set the seed and generate the same
+        # num_tasks random permutations for both training and validation datasets; the
+        # first one is the identity permutation (i.e., regular MNIST), represented
+        # below as `None`
+        g = torch.manual_seed(seed)
 
-        self.permutations = PermutedMNIST.permutations
+        self.permutations = [
+            torch.randperm(784, generator=g) for task_id in range(1, num_tasks)
+        ]
+        self.permutations.insert(0, None)
 
     def __getitem__(self, index):
         """
@@ -103,22 +98,14 @@ class ContextDependentPermutedMNIST(PermutedMNIST):
     vector along with the data sample and target.
     """
 
-    # Context vectors should be the same whether this class is instantiated for
-    # training or testing
-    contexts = None
+    def __init__(self, num_tasks, dim_context, seed, train, root=".",
+                 target_transform=None, download=False):
 
-    def __init__(self, num_tasks, dim_context, train, root=".", target_transform=None,
-                 download=False):
+        super().__init__(num_tasks, seed, train, root, target_transform, download)
+        self.dim_context = dim_context
 
-        super().__init__(num_tasks, train, root, target_transform, download)
-
-        # Generate context vectors if `contexts` isn't already defined
-        if ContextDependentPermutedMNIST.contexts is None:
-            ContextDependentPermutedMNIST.contexts = generate_context_vectors(
-                num_contexts=num_tasks, n_dim=dim_context, percent_on=0.05
-            )
-
-        self.contexts = ContextDependentPermutedMNIST.contexts
+        # Initialize random binary sparse context vectors for each permutation
+        self.init_contexts(seed)
 
     def __getitem__(self, index):
         """
@@ -127,6 +114,26 @@ class ContextDependentPermutedMNIST(PermutedMNIST):
         img, target = super().__getitem__(index)
         task_id = self.get_task_id(index)
         return img, self.contexts[task_id, :], target
+
+    def init_contexts(self, seed):
+        percent_on = 0.05
+        num_contexts = self.num_tasks
+
+        num_ones = int(percent_on * self.dim_context)
+        num_zeros = self.dim_context - num_ones
+
+        self.contexts = torch.cat((
+            torch.zeros((num_contexts, num_zeros)),
+            torch.ones((num_contexts, num_ones))
+        ), dim=1)
+
+        # Shuffle each context vector i.i.d. to randomize it; use a fixed seed during
+        # shuffling so that train & validation contexts are identical
+        g = torch.manual_seed(seed)
+
+        for i in range(num_contexts):
+            self.contexts[i, :] = self.contexts[i, torch.randperm(self.dim_context,
+                                                                  generator=g)]
 
 
 def permute(x, permutation):
--- a/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
+++ b/nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
@@ -1,0 +1,145 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2021, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


subutai:
@akashvelu There are some changes to dendritic_mlp you might want to look at as well - might conflict with your changes.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        self.contexts = ContextDependentPermutedMNIST.contexts
+
+    def __getitem__(self, index):
+        """
+        Returns an (image, context, target) triplet.
+        """
+        img, target = super().__getitem__(index)
+        task_id = index // len(self.data)
```
subutai, line 127: What about adding a get_task_id(index) method in PermutedMNIST? It would return an int and might be generally useful. Then this line (and also line 78) should use that method.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
```diff
@@ -0,0 +1,145 @@
...
+        return img, self.contexts[task_id, :], target
```
subutai, line 128: Maybe also return `task_id` here?

projects/dendrites/permutedMNIST/run_dendritic_network.py
```diff
@@ -0,0 +1,231 @@
...
+        num_classes_per_task = math.floor(num_classes / num_tasks)
+
+        task_indices = defaultdict(list)
+        for i in range(num_tasks):
+            for j in range(num_classes_per_task):
+                task_indices[i].extend(class_indices[j + (i * num_classes_per_task)])
+        return task_indices
+
+    def run_iteration(self):
+        # Reset optimizer before training on a new task, just as context-dependent
```
subutai, line 75: Is this ever called in the script? I couldn't see where.

subutai (approved):
Looks good. Left a couple of comments and suggestions.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: What's the need? If the context vector is provided, then that itself is task information. If we return `task_id`, then it should be eventually used to select the appropriate context vector, which means contexts would be separate from the dataset class. That would make sense, but it still seems better to do it this way.

subutai: Ok, makes sense.

projects/dendrites/permutedMNIST/run_dendritic_network.py
karangrewal: I removed it, and placed the optimizer reset within the continual learning loop.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: Added.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
```diff
@@ -0,0 +1,146 @@
...
+        `index`, determines which pixel-wise permutation to apply. Target values are
+        also scaled to be unique to each permutation.
+        """
+        img, target = super().__getitem__(index % len(self.data))
+
+        # Determine which task `index` corresponds to
+        task_id = self.get_task_id(index)
+
+        # Apply permutation to `img`
+        img = permute(img, self.permutations[task_id])
```
lucasosouza, line 81: where is `permute` coming from?

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
```diff
@@ -0,0 +1,146 @@
...
+        ])
+        super().__init__(root=root, train=train, transform=data_transform,
+                         target_transform=target_transform, download=download)
+
+        self.num_tasks = num_tasks
+
+        # Generate num_tasks random permutations - the first one is the identity
+        # permutation (i.e., regular MNIST), represented below as `None`
+        if PermutedMNIST.permutations is None:
+            PermutedMNIST.permutations = [
```
lucasosouza, line 60: `permutations` is clearly an instance variable, since it is modified in the init method of a class (the init initializes an instance). Why is it defined as a class variable?

lucasosouza, line 60: `permutations` is clearly an instance variable, since it is modified in the init method of a class (the init initializes an instance). Why is it defined as a class variable?

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
```diff
@@ -0,0 +1,146 @@
...
+    contexts = None
+
+    def __init__(self, num_tasks, dim_context, train, root=".", target_transform=None,
+                 download=False):
+
+        super().__init__(num_tasks, train, root, target_transform, download)
+
+        # Generate context vectors if `contexts` isn't already defined
+        if ContextDependentPermutedMNIST.contexts is None:
+            ContextDependentPermutedMNIST.contexts = generate_context_vectors(
```
lucasosouza, line 117: Same coment applues here, why is `contexts` a class variable

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
lucasosouza: I don't think modifying the signature of dataloader getitem is a good alternative. Dataloader is just data and target, and this case your data happens to be not just an image, but an image but a context. It could potentially be a composition of even more things.

Python is not type enforced and type hints are commonly used with unions of types, that flexibility exists by design. So data and target could be of different types. I think the best alternative here is returning a tuple data (img, context) and target, preserving the original signature.

projects/dendrites/permutedMNIST/run_dendritic_network.py
```diff
@@ -0,0 +1,205 @@
...
+from nupic.research.frameworks.vernon import ContinualLearningExperiment, mixins
+
+
+# ------ Experiment class
+class PermutedMNISTExperiment(mixins.RezeroWeights,
+                              ContinualLearningExperiment):
+    pass
+
+
+# ------ Training & evaluation functions
```
lucasosouza, line 47: Are all dendrites experiment still done through this loop? As a suggestion, this could be rewritten in a few lines of code if using Vernon.

lucasosouza (commented):
I will mark as comment so you can decide whether to implement or not.

nupic/research/frameworks/vernon/experiments/cl_experiment.py
```diff
@@ -138,7 +138,8 @@ def create_validation_sampler(cls, config, dataset):
     def compute_task_indices(cls, config, dataset):
         # Assume dataloaders are already created
         class_indices = defaultdict(list)
-        for idx, (_, target) in enumerate(dataset):
+        for idx, batch_item in enumerate(dataset):
```
lucasosouza, line 141: See comment on preserving the dataloader signature. It is a small change and won't really modify things a lot here, just how you unpack.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
lucasosouza: I don't think modifying the signature of dataloader getitem is a good alternative. Dataloader is just data and target, and this case your data happens to be not just an image, but an image but a context. It could potentially be a composition of even more things.

Python is not type enforced and type hints are commonly used with unions of types, that flexibility exists by design. So data and target could be of different types. I think the best alternative here is returning a tuple data (img, context) and target, preserving the original signature.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: Since we instantiate a `PermutedMNIST` object twice (once for training and once for validation), we need the same pixel-wise permutations applied to images for each task. I found that defining `permutations` as a class variable is the easiest way to accomplish this.

One other way to accomplish the same would be to make `permutations` an optional argument to the `__init__` method, and that way when the validation class is instantiated, we can pass in the training class's permutations so that the same pixel-wise permutations are also applied to the validation examples. However, this may get messy since I'm not sure how the configs in Vernon would do this exactly.

Do you have any preferences on which way to go?

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: Same reason why `permutations` is a class variable.

projects/dendrites/permutedMNIST/run_dendritic_network.py
karangrewal: Yes. I'm planning to merge everything with Vernon soon, but for now, there are slight modifications I needed to make to the training & evaluation functions that's easier done if I just redefine the functions themselves.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
lucasosouza: There are several ways of solving this, but I don't think class attributes is one of them. Even though it might work in one tested used case, that is not what class attributes are for and can cause hidden bugs. 

Training and validation could share the dataloader. If they can't, and dataloader has to share permutations, then composition makes sense - permutations belonging to a higher level class (like Experiments), and passed as an object to dataloader. 

But the easiest way to solve is just seed it. Have a seed be initialized in the higher class and be passed as an argument to the dataloader, no need to pass all permutations. This will always yield the same results: 
```
g = torch.Generator()
g.manual_seed(30)
torch.randperm(100, generator=g)
```

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
lucasosouza: @karangrewal just making sure you saw this comment as well

lucasosouza:
@karangrewal please see comment on dataloader getitem signature, I think it went it unnoticed in the review because I added the comment to an existing thread

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: @lucasosouza noted! I can make this change so that the dataloader yields ((data, context), target) tuples.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: I think this is a good suggestion. I modified the dataset class so that `seed` is now an argument (specified in the config), and that's used to generate the permutations and contexts.

nupic/research/frameworks/pytorch/datasets/permuted_mnist.py
karangrewal: Fixed.

lucasosouza (approved):

PR #514, opened 2021-05-13
Wide BERT and Getting the Full 80% Sparsity

mvacaporale:
This PR includes experiments to train wider versions of BERT and those the get the full 80% sparsity.

Here's a table of results with Wide BERT
| width | sparsity |on-params | eval loss |
|:-----:|:---------|:---------:|:-:|
| 1.0x | 80%    | 850,510 |3.578|
| 1.25x | 84.3% | 842,474 | 3.514 |
| 1.50x | 88%   | 865,227 | 3.461 |
| 2.0x | 90.83%  | 843,781 | 3.469 |
| 4.875x | 97.2%  | 834,243 | 3.438 |
| 4.875x (8 att. heads) | 97%  | 918,441 | 3.409 |
| Small BERT | 96.95% | 924,208 | 3.317 |

The wider architecture does seem to help, although given a comparable number if params, is doesn't do any better than Small BERT with a deeper architecture. Thus, there's still more to investigate.

Per the full 80% sparsity: There's a new param `sparsify_all_embeddings`. That sparsify's the word embeddings, as usual, but also the token and position embeddings. This is now the default, and old configs have been updated accordingly. Results show that sparsifying these  additional embeddings actually helps accuracy on both pre-training and fine-tuning. As well, since the model has dense layer normalization params, we'll need to set the sparsity slightly higher than 80%. In the config, that will look something like `sparsity=0.801`.


benja-matic:
Nice work, looks good to me. Duly noted re new param sparsify_all_embeddings.

projects/transformers/README.md
```diff
@@ -35,6 +35,9 @@ Known issues, to be investigated/fixed:
 
 > Finetuning is important to evaluate a language model effectiveness when being used for transfer learning in downstream tasks. However, it is not a reliable metric to use for optimization. These numbers will vary with a significant margin of error between two different runs. Use perplexity instead to guide hyperparameter search and development of language models.
 
+Known Bugs
```
lucasosouza, line 38: I haven't observed this before, and I ran dozens of finetuning runs,so seems like a bug introduced by a recent version. Is there a way to set wandb logging default to False in finetuning, so we avoid this? Most of the finetuning runs we only care about the end result.

Also, do you believe this will persist once we introduce deepspeed?

lucasosouza (approved):
Looks good to me. About the impact sparsifying token and position embedings, Subutai made a good point that we should make sure these results extend to full BERT as well, not just bertitos.

projects/transformers/README.md
mvacaporale: I believe one sets `WANDB_DISABLED=false`. Are you suggesting it to be set automatically? I'm curious if the problem will persist. It happened to me only one day, so maybe I had did something wrong. I think if you and Ben encounter it, then maybe we should disable wandb automatically.

mvacaporale: I'm not sure how it will interact with DeepSpeed.

projects/transformers/experiments/__init__.py
```diff
@@ -28,6 +28,7 @@
 from .bert_replication import CONFIGS as BERT_REPLICATION
 from .bertitos import CONFIGS as BERTITOS
 from .distillation import CONFIGS as DISTILLATION
+from .eighty_percent_sparse import CONFIGS as EIGHT_PERCENT_SPARSE
```
benja-matic, line 31: nitpick, but any reason all capitals is "EIGHT" not "EIGHTY"?

benja-matic (approved):
(Actually approving this time)

PR #515, opened 2021-05-17
Centroid method for inferring context signal

karangrewal:
This PR introduces

- the `CentroidContext` mixin to train dendritic networks that use the centroid/prototype method to (a) construct a prototype context signal for training, and (b) infer a previously-seen context during inference,
- the `PermutedMNISTTaskIndices` mixin that is only compatible with permutedMNIST experiments, and significantly speeds up initialization since no data transformations are applied when task indices are computed, and
- (minor) new configs added to the `projects/dendrites/permutedMNIST` folder.

The train & evaluate functions in dendrite model utils have also been updated to accommodate the centroid setting.


nupic/research/frameworks/vernon/mixins/centroid_context.py
```diff
@@ -0,0 +1,128 @@
...
+        if isinstance(x, list):
+            x = x[0]
+        x = x.flatten(start_dim=1)
+        n_x = x.size(0)
+
+        centroid_vector = centroid_vector.to(x.device)
+
+        centroid_vector = n_centroid * centroid_vector + n_x * x.mean(dim=0)
+        centroid_vector = centroid_vector / (n_centroid + n_x)
+        n_centroid += n_x
```
subutai, line 110: This seems a bit convoluted. We just want the mean vector, right? 

Why not just:
`centroid_vector += n_x*x.mean(dim=0)`
in the inner loop, and then after the loop:
`centroid_vector = centroid_vector / n_centroid`

nupic/research/frameworks/vernon/mixins/centroid_context.py
```diff
@@ -0,0 +1,128 @@
...
+def infer_centroid(contexts):
+    """
+    Returns a function that takes a batch of test examples and returns a 2D array where
+    row i gives the the centroid vector closest to the ith test example.
+    """
+
+    def _infer_centroid(data):
+        context = torch.cdist(contexts, data)
+        context = context.argmin(dim=0)
+        context = context.cpu()
```
subutai, line 124: Why `.cpu()` here?

projects/dendrites/permutedMNIST/experiments/base.py
```diff
@@ -93,8 +106,170 @@ class PermutedMNISTExperiment(mixins.RezeroWeights,
...
+    num_classes=500,  # 10 * NUM_TASKS
+
+    optimizer_class=torch.optim.Adam,
+    optimizer_args=dict(
+        lr=tune.sample_from(lambda spec: np.random.choice([0.001, 0.0005])),
+    ),
+)
+
+# Centroid method for inferring contexts: 10 permutedMNIST tasks
+CENTROID_10 = dict(
```
subutai, line 211: I think we should move all the basic centroid experiments into its own experiments/centroid.py file.

subutai (changes requested):

projects/dendrites/permutedMNIST/experiments/base.py
karangrewal: Done.

nupic/research/frameworks/vernon/mixins/centroid_context.py
karangrewal: I think this is leftover from some debugging I did weeks ago. I removed it, since it seems like everything runs without error if we omit the cpu call.

nupic/research/frameworks/vernon/mixins/centroid_context.py
karangrewal: Done.

nupic/research/frameworks/vernon/mixins/centroid_context.py
```diff
@@ -0,0 +1,128 @@
...
+
+class CentroidContext(metaclass=abc.ABCMeta):
+    """
+    When training a dendritic network, use the centroid method for computing context
+    vectors (that dendrites receive as input) for both training and inference.
+    """
+
+    def setup_experiment(self, config):
+        super().setup_experiment(config)
+
```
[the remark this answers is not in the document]
subutai: Probably a good idea.

subutai (approved):

nupic/research/frameworks/vernon/mixins/__init__.py
```diff
@@ -42,6 +43,7 @@
 from .maxup import MaxupPerSample, MaxupStandard
 from .multi_cycle_lr import MultiCycleLR
 from .oml import OnlineMetaLearning
+from .permuted_mnist_task_indices import *
```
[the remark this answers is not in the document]
karangrewal: This is the same, since in the permuted_mnist_task_indices.py file, I defined `*` to be

```
__all__ = [
    "PermutedMNISTTaskIndices",
]
```

so `import *` will just import the class.

nupic/research/frameworks/vernon/mixins/__init__.py
```diff
@@ -19,6 +19,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+from .centroid_context import *
```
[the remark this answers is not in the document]
karangrewal: same reason as below.

nupic/research/frameworks/vernon/mixins/__init__.py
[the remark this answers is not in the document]
jeremyforest: ah ok got it :+1:

nupic/research/frameworks/vernon/mixins/centroid_context.py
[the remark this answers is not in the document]
karangrewal: Done.

jeremyforest (approved):

PR #521, opened 2021-05-27
Better drop_last fix, updated results table (RES-2222)

benja-matic:
Replaced the old drop_last fix with a more durable solution, which overrides the predict and evaluate methods of the trainer. This fix covers the cases where those methods are called internally in HF code, not just by us. 

Updated the readme results table: bert100k, bert_1mi, and sparse_80%_kd_onecycle_lr_rigl. The bert_1mi results are calculated using an extra 10 runs on wnli task.

There are some new experiments I'm playing with, like the "simple but hard to beat" baseline in finetuning.py. That stuff might still move around. Wanted to get this PR going for the results and the new fix.


mvacaporale:
Have you tried running with `tiny_bert_linear_lr_range_test`? It would be good to know this can fully run without any issues as the lr-range tests validate every epoch.

projects/transformers/README.md
```diff
@@ -9,9 +9,9 @@ In progress, current results:
...
-| bert_1mi |          80.76 |          76.82 |  48.87 | 84.08/84.57 | 89.76/85.68 |  91.19 | 90.58/87.17 | 66.02 |  91.44 | 87.67/87.54 |  45.31 | 5.013 | 1.612 |
-| bert_100k |          75.71 |          72.51 |  40.98 | 78.26/78.65 | 84.05/77.86 |  87.74 | 89.12/85.25 |  58.2 |  88.31 | 83.90/83.80 |  46.88 | 8.619 | 2.154 |
-| sparse_80%_kd_onecycle_lr_rigl |            75.3 |          72.83 |  38.29 | 79.72/80.88 | 87.31/82.81 |  87.65 | 89.19/85.25 |  54.3 |  88.89 | 80.74/80.59 |  53.12 | 8.482 | 2.138 |
+| bert_1mi |          80.13 |          77.17 |  45.81 | 84.27/84.63 | 88.26/83.82 |  91.21 | 90.54/87.20 | 65.34 |  91.86 | 87.41/87.43 |  53.52 | 5.013 | 1.612 |
+| bert_100k |          75.36 |          71.68 |  39.56 | 78.88/79.08 | 82.71/76.23 |  87.77 | 89.31/85.57 |  58.12 |  87.61 | 83.95/83.84 |  42.25 | 8.619 | 2.154 |
+| sparse_80%_kd_onecycle_lr_rigl |            75.3 |          72.57 |  36.49 | 79.23/79.66 | 86.55/81.86 |  88.23 | 89.39/85.64 |  54.51 |  90.6 | 81.31/81.24 |  50.7 | 8.482 | 2.138 |
```
mvacaporale, line 14: These updated results are without the simple but hard to beat baseline, correct?

projects/transformers/experiments/finetuning.py
```diff
@@ -129,6 +129,38 @@
     model_name_or_path="/mnt/efs/results/pretrained-models/transformers-local/bert_100k",  # noqa: E501
 )
 
+# the name 'simple' is in reference to the paper
```
mvacaporale, line 132: This may be more of a nit pick style comment, but I believe in using proper grammar and punctuation in comments, especially longer ones. I believe it makes it easier to read. This one could be formatted as 
```
# The name 'simple' is in reference to the paper "On the stability of finetuning BERT"
# where they propose a "simple but hard to beat" approach
#       https://openreview.net/pdf?id=nzpLWnVAyah
# 
# How to decided num_train_epochs: They say rte for 20 epochs is good ...
```
Notice how I try and use the full line length as well. There should be an extension in vscode to help automatically format comments for line length.

projects/transformers/run_utils.py
```diff
@@ -708,6 +676,22 @@ def init_model(model_args, config, tokenizer, finetuning=False):
     return model
 
 
+def toggle_drop_last_factory(method):
```
mvacaporale, line 679: I believe this method is described as a wrapper than a factory. Would you agree?

mvacaporale: I like this solution btw. Seems like a good way to modify the function without having to create a new subclass to Trainer.

mvacaporale (approved):
Looks good. Thanks for catching the error with the first fix. Approved barring
* some nit pick comments about grammar and punctuation,
* and confirmation that this fix work with the lr-range test.

lucasosouza (approved):
Looks great

projects/transformers/run_utils.py
```diff
@@ -708,6 +676,23 @@ def init_model(model_args, config, tokenizer, finetuning=False):
     return model
 
 
+def toggle_drop_last_wrapper(method):
+    """
+    Return a function that turns drop_last off before it is called. Used for ensuring
+        trainer.args.dataloader_drop_last is False during evaluation steps. After
```
mvacaporale, line 682: It seems you accidentally added an indent here in the comments.

PR #522, opened 2021-05-28
Add deepspeed version of LR range test experiment

lscheinkman:
This lr-range test is based on `bert_sparse_100k_kd_lr_range_test` and adapted for deepspeed training.
With this test the best `max_lr` value found is `0.0017`

### DIFF
--- a/projects/transformers/experiments/ablations.py
+++ b/projects/transformers/experiments/ablations.py
@@ -290,9 +290,9 @@ finetuning_bert_sparse_kd_oncycle_lr_100k_glue.update(
     model_name_or_path="/mnt/efs/results/pretrained-models/transformers-local/bert_sparse_80%_kd_onecycle_lr_100k",  # noqa: E501
 )
 
-#############
-# Deepspeed #
-#############
+# ---------
+# Deepspeed
+# ---------
 
 # This lr-range test is based on `bert_sparse_100k_kd_lr_range_test` and adapted
 # for deepspeed training.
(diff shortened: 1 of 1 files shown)


projects/transformers/experiments/ablations.py
```diff
@@ -291,6 +290,42 @@ def pct_start_hp_space(trial):
     model_name_or_path="/mnt/efs/results/pretrained-models/transformers-local/bert_sparse_80%_kd_onecycle_lr_100k",  # noqa: E501
 )
 
+#############
+# Deepspeed #
+#############
```
mvacaporale, line 295: Nit pick: For consistency with the other files, could you format the section header by
```
# ------------
# Deepspeed
# ------------
```

projects/transformers/experiments/ablations.py
```diff
@@ -291,6 +290,42 @@ def pct_start_hp_space(trial):
...
+
+# This lr-range test is based on `bert_sparse_100k_kd_lr_range_test` and adapted
+# for deepspeed training.
+# With this test the best `max_lr` value found is `0.0017`.
```
mvacaporale, line 299: Didn't the max_lr of 0.0012 cause a nan loss? I wonder why this test is suggesting something higher.

mvacaporale (approved):
Looks good. Thanks for figuring out how to run the lr-range test with Deepspeed.

projects/transformers/experiments/ablations.py
lscheinkman: the nan loss was fixed with `"gradient_clipping": 1.0,`

PR #524, opened 2021-06-04
Vernon dendrites 3

karangrewal:
This pull request contains scripts for

- generating figures used in the CNS submission, and
- generating context vectors for dendritic networks via a randomly-initialized and untrained spatial pooler.


subutai (commented):
I don't think we should have xy_10_kxtmyf.zip in master. If it's just a few datapoints, best to just stick them as lists in the notebook itself.

projects/dendrites/permutedMNIST/centroid_for_figure_1b.py
```diff
@@ -0,0 +1,286 @@
+#  Numenta Platform for Intelligent Computing (NuPIC)
+#  Copyright (C) 2021, Numenta, Inc.  Unless you have an agreement
```
subutai, line 2: Why have a whole complete source file for one figure? Can't we just run existing code with specific parameters? Or is there something special about this version?

subutai (changes requested):

projects/dendrites/permutedMNIST/centroid_for_figure_1b.py
karangrewal: We use the `PlotHiddenActivations` mixin to create that figure, but the mixin relies on target values. The problem here is that if `share_labels = True`, then the labels for examples in different tasks overlap, so we couldn't generate the figure. I only modification I made here is to have a separate function for evaluation that doesn't squish all labels into {0, 1, ..., 9}, so the plots can be created properly, which is called `tracking_eval`.

karangrewal: I agree that it seems wasteful to have a whole script just for this. I can modify the original script so that it has a "2nd evaluation loop" specifically for plotting. What do you think?

nupic/research/frameworks/dendrites/mixins/sp_context.py
```diff
@@ -0,0 +1,99 @@
...
+
+class SpatialPoolerContext(metaclass=abc.ABCMeta):
+    """
+    Use an untrained SP to output context vectors.
+    """
+
+    def setup_experiment(self, config):
+        context_model_args = config.get("context_model_args")
+        context_out = context_model_args.get("output_size")
+        dim_context = config["model_args"].get("dim_context")
```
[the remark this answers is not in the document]
subutai: Good point. I added these in, which in turn made me simplify the way arguments are handled.

projects/dendrites/permutedMNIST/centroid_for_figure_1b.py
subutai: Ok, sounds good.

karangrewal: I removed the script, and added a mixin that runs the additional eval loop using **task labels** instead of **shared target labels** (which is exactly what we want for plotting).

nupic/research/frameworks/vernon/mixins/centroid_figure_1b.py
```diff
@@ -0,0 +1,111 @@
+#  Numenta Platform for Intelligent Computing (NuPIC)
```
subutai, line 1: Can you move this to dendrites/mixins? I don't think such a specific file should be in the Vernon directory structure.

subutai (changes requested):

nupic/research/frameworks/vernon/mixins/centroid_figure_1b.py
karangrewal: Done.

subutai: Looks like the init files and imports were changed, but the file itself wasn't moved in the commit.

karangrewal: Are you sure? It tells me the file was indeed moved. Also,

- [dendrites/mixins/](https://github.com/karangrewal/nupic.research/tree/vernon-dendrites-3/nupic/research/frameworks/dendrites/mixins) folder on this branch,
- [vernon/mixins folder](https://github.com/karangrewal/nupic.research/tree/vernon-dendrites-3/nupic/research/frameworks/vernon/mixins) on this branch

and only the latter contains the file `centroid_figure_1b.py`.

projects/dendrites/permutedMNIST/centroid_for_figure_1b.py
subutai: 👍

nupic/research/frameworks/vernon/mixins/centroid_figure_1b.py
subutai: Yes, you're right. I think my various github tabs did not update in sync, even after a refresh. Sorry about that.

subutai (approved):

PR #525, opened 2021-06-07
Track evaluation metrics throughout training (RES-2234)

benja-matic:
- I added TrackEvalMetrics callback so every time trainer.evaluate() is called, the callback stores evaluation metrics. 
- I set the default callbacks in finetuning.py to [TrackEvalMetrics, EarlyStoppingCallback].
- I adjusted TaskResults class in run_utils.py to handle combinations of the above callbacks while staying backward compatible. If TrackEvalMetrics, then the reduce method involves reducing over runs, and over evaluation calls within a run.
- Early stopping requires load_best_model_at_end, metric_for_best_model, evaluation_strategy, so I updated the finetuning configs to have defaults for these. Eval_steps and early_stopping patience should be set manually.
- export_finetuning_results.py works as before, but if you unpickle the task_results.p file, you'll have access to the time series of evaluation metrics.
- In the special case on mnli finetuning, there are two eval sets. Primitive way to deal with this for now is to track the "matched" eval set metrics during training, and evaluated the "mismatched" set once at the end of training. The metric_key_prefix argument is set to "mm" in this case, so TrackEvalMetrics can recognize this one case and handle with a separate data structure. If early stopping is enabled, note that load_best_model_at_end is also enabled, so the correct model will be evaluated on the mismatched set either way.


projects/transformers/run_utils.py
```diff
@@ -825,28 +840,84 @@ class TaskResults():
...
     def reduce_metrics(self, reduction="mean"):
+        """
+        Get average or max over runs. Handles two cases:
+            1) If using early stopping, find the index where metric_for_best_model
+               is highest in each run. Get the max or mean of these.
+            2) If not using early stopping, get the last entry in each run, and take
+               the max or mean of these. This entry corresponds to the end of training.
+
+        Note that if you're not using TrackEvalMetrics callback, results dictionaries
+        are formatted so the values are lists fo length 1. Cases (1) and (2) above
```
mvacaporale, line 887: Typo "of" instead of "fo"

projects/transformers/run_utils.py
```diff
@@ -693,6 +693,20 @@ def toggle_method(*args, **kwargs):
     return toggle_method
 
 
+def format_eval_results(eval_results, run, task_name):
```
mvacaporale, line 696: Maybe include an example of what this outputs in the docstring for clarity.

mvacaporale (approved):
Looks good. Only left small comments and questions.

projects/transformers/callbacks/metrics.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Using a workaround for mnli, and only evaluating mnli-mm at the very
+        # end of training. Therefore don't update anything
+        is_mnli_mm = False
+        for key in metrics.keys():
+            if key == "mm_accuracy":
+                is_mnli_mm = True
+                break
+
+        if is_mnli_mm:
```
lucasosouza, line 81: Another suggestion, to separate the loop and the transformation. Entirely up to you, just giving some suggestions to separate concerns and simplify the coding (lambda could also be a def for easier reading):
```
split = lambda k: k.split("_", 1)[1] if "_" in k else k
mm_metrics = {split(k): v for k, v in metrics.items()}
```

edit: if you end up reusing that `split` function above in other files, it could merit being a separate function altogether (with a more descriptive name like `handle_composite_metrics`). So if there is any maintenance required to handle a new metric in the future, all we would have to do is change that function.

projects/transformers/experiments/finetuning.py
```diff
@@ -37,10 +41,48 @@
 https://wandb.ai/numenta/huggingface/reports/Static-Sparse-Baselines--Vmlldzo1MTY1MTc
 """
 
+# Runs can easily break because metric_for_best_model varies by task,
+# but is required if using early_stopping. So it's easy to specify a metric
+# that isn't present for the current task. This is a reference to help avoid that.
```
lucasosouza, line 46: Should this be an assert somewhere in the callback?

projects/transformers/experiments/finetuning.py
```diff
@@ -51,10 +93,19 @@
...
     do_eval=True,
     do_predict=True,
+    eval_steps=10,
+    evaluation_strategy="steps",
+    load_best_model_at_end=True,
     per_device_train_batch_size=32,
     per_device_eval_batch_size=32,
     learning_rate=2e-5,
-    num_train_epochs=3,
+    warmup_ratio=0.1,
```
lucasosouza, line 102: No action here, just a comment: that is the first I see it defined as a warmup_ratio, that is interesting. Previously I've only seen it defined as an absolute number (warmup_steps) in HF examples.

projects/transformers/experiments/finetuning.py
```diff
@@ -78,16 +129,18 @@
...
     override_finetuning_results=False,
     task_hyperparams=dict(
-        wnli=dict(num_runs=2, learning_rate=2e-4),
-        rte=dict(num_runs=0),
+        wnli=dict(num_runs=2, max_steps=20, learning_rate=2e-4),
+        rte=dict(num_runs=1),
         cola=dict(num_runs=2),
-        stsb=dict(num_runs=1),
+        stsb=dict(num_runs=1,
+                  metric_for_best_model="pearson"),
```
lucasosouza, line 142: Do we have to define for each task? Is there a default set somewhere, so if we just say use_best_model it will automatically pick up that metric?

projects/transformers/experiments/finetuning.py
```diff
@@ -108,9 +161,13 @@
     do_train=True,
     do_eval=True,
     do_predict=False,
+    eval_steps=10,
+    evaluation_strategy="steps",
+    load_best_model_at_end=True,
```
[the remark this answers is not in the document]
lucasosouza: I might be wrong, but the goal is to evaluate at the end using the best model instead of using the model at the last step (which as we've seen in some finetuning tasks it is usually not the model at the last step due to overfitting).

projects/transformers/experiments/finetuning.py
```diff
@@ -133,39 +194,48 @@
...
 
 finetuning_bert100k_glue_simple = deepcopy(finetuning_bert100k_glue)
 finetuning_bert100k_glue_simple.update(
-    warmup_ratio=0.1,
+    run_name="finetuning_bert100k_glue_simple",
     task_hyperparams=dict(
-        cola=dict(num_train_epochs=6, num_runs=5),  # 6 * 8500 ~ 50k
+
+        cola=dict(eval_steps=50,
+                  max_steps=1562,
```
lucasosouza, line 218: Could this calculation be donen automatically via a function, given there is a defined heuristic which is given in the comment above? That would avoid having the user calculate it when creating the config. The fewer human interactions the better, not just for efficiency but to avoid accidental data entry mistakes.

projects/transformers/experiments/trifecta.py
```diff
@@ -265,6 +265,7 @@ class KDLRRangeTestTrainer(LRRangeTestMixin,
     model_type="fully_static_sparse_bert",
     model_name_or_path="/mnt/efs/results/pretrained-models/transformers-local/"
     "bert_sparse_80%_trifecta_100k",
+    run_name="finetuning_bert_sparse_trifecta_100k_glue_simple"
```
lucasosouza, line 268: There is a rule in run.py (line 114) that if `run_name` is not given, it will automatically get the config name, so this would be redundant.

projects/transformers/run.py
```diff
@@ -334,9 +354,42 @@ def run_finetuning_single_task(
...
+            )
+            n_evals = len(tracked_eval_metrics["steps"])
+            mm_dict = metric_callback.mm_metrics
+            for key in mm_dict.keys():
+                key_name = "mm_eval_" + key
+                # Fill a list of same length as other metrics for consistency
+                tracked_eval_metrics[key_name] = [mm_dict[key] for i in range(n_evals)]
+
+        eval_results = tracked_eval_metrics
+
```
lucasosouza, line 390: The entire block above is getting super long, we could move this into a separate function that gets called (or pieces of it at least, like how to handle the mnli exception)

projects/transformers/run.py
```diff
@@ -408,15 +461,26 @@ def run_finetuning_multiple_tasks(
...
+        task_results = TaskResults(task_name,
+                                   has_early_stopping,
+                                   training_args=training_args)
 
         # Run finetuning and save results
         for _ in range(training_args.num_runs):
             # reset seed per run
-            training_args.seed = random.randint(0, 1000000000)
+            training_args.seed = random.randint(0, 1_000_000_000)
```
lucasosouza, line 479: Huh, interesting, I had no idea 1_000_000_000 was a thing in Python. A lot easier to read.

projects/transformers/run_utils.py
```diff
@@ -693,6 +693,20 @@ def toggle_method(*args, **kwargs):
...
+
+    ignore_keys = ["steps", "epoch"]
+    new_eval_results = {}
+    for key in eval_results.keys():
```
lucasosouza, line 700: That same approach discussed in the callback could be used to simplify this code block

lucasosouza (approved):
It looks good, I didn't find any thing that could be a bug or a major imcompatibility with existing code, hence tagging as approved so we can move along.

However, I did left a number of comments on things I believe it can be improved, I would recommend going over them if you have the time. In general, known heuristics could be automated to facilitate usage by other researchers and avoid bugs caused by human errors when defining a config.

projects/transformers/experiments/finetuning.py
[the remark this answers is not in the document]
benja-matic: It looks for the checkpoint with the best metric_for_best_model, and loads that model after training, in this case on a finetuning task. It uses that model (as opposed to one saved at a later checkpoint) for evaluation or prediction steps next. load_best_model_at_end is also required if you are using EarylStoppingCallback.

projects/transformers/experiments/finetuning.py
```diff
@@ -133,38 +194,35 @@
...
 #
-#       elif 50k <= len(train_dataset) < 300k
+#       else
 #           use the default of 3 epochs
-#           (sst2, wnli)
 #
-#       elif len(train_dataset) >= 300k
-#           train for 1 epoch
-#           (qqp, mnli)
+# Note that EarlyStoppingCallback is in use, which was not mentioned in the paper
```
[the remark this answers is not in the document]
benja-matic: The original reason was just to do the same thing as load_best_model_at_end. I'll check if you can do that without early stopping. Early stopping is potentially useful nevertheless for saving time on failed runs. It also makes it so you can set max_steps to a large number, with less concern about overfitting.

projects/transformers/run.py
```diff
@@ -334,9 +354,42 @@ def run_finetuning_single_task(
...
+    # TrackEvalMetrics callback instance. This means you need to know its index
+    # in the model_args.trainer_callbacks list. This block finds it.
+    tracked_metrics = False
+    tracked_metrics_idx = None
+    for callback_idx in range(len(model_args.trainer_callbacks)):
+        if isinstance(model_args.trainer_callbacks[callback_idx], TrackEvalMetrics):
+            tracked_metrics = True
+            tracked_metrics_idx = callback_idx
+
+    # If already tracking metrics, no need to evaluate again at the very end
```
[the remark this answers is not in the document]
benja-matic: Yes, good catch. If using load_best_model_at_end though, it would be redundant. I'll update this block to handle these cases properly, and also move it to a separate function as Lucas suggested.

projects/transformers/callbacks/metrics.py
```diff
@@ -0,0 +1,90 @@
...
+        self.train_metrics = {}  # ignore since this could slow down training
+        self.steps = []
+        self.step_counter = 0
+
+    def on_evaluate(self, args, state, control, metrics, **kwargs):
+        """Update steps counter, training metrics, & eval metrics"""
+
+        is_mnli_mm = self.check_for_mnli(metrics)
+        if not is_mnli_mm:
+            self.step_counter += args.eval_steps
```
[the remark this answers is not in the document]
benja-matic: Model gets evaluated every eval_steps. These results later get stored in task_results.p. This is so that when you load that file, you can also select a run, and plot loss vs. training step.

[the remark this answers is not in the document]
lucasosouza: Still not clear to me. Isn't evaluation just a single pass over the eval dataset, with no gradient steps? And if there are no gradient steps in eval, why do we need to track intermediate evaluation results?

projects/transformers/constants.py
```diff
@@ -0,0 +1,61 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the GNU Affero Public License for more details.
+#
+#  You should have received a copy of the GNU Affero Public License
+#  along with this program.  If not, see htt"://www.gnu.org/licenses.
+#
+#  http://numenta.org/licenses/
+#
+
+"""
```
lucasosouza, line 21: These all look good, but they are related to finetuning - maybe for now we can make this a finetuning specific file and rename it accordingly.

projects/transformers/export_finetuning_results.py
```diff
@@ -32,17 +32,235 @@
...
+        if save_name:
+            plt.savefig(save_name)
+
+        print("Figure ready, hit plt.show() to visualize")
+
+        return fig, ax
+
+    """
+    No quality assurance or testing below this line within this class. These
+    methods are being rapidly prototyped and will be revised greddily as
```
lucasosouza, line 132: typo: greedily

lucasosouza (approved):
These look good. Left two small comments. We should go ahead and merge it after the remaining small fixes pending.

PR #526, opened 2021-06-07
RES-2225: Few optimizations 

lscheinkman:
@lucasosouza Please review.
These are few optimization discovered while debugging deepspeed integration

### DIFF
--- a/projects/transformers/experiments/base.py
+++ b/projects/transformers/experiments/base.py
@@ -46,7 +46,8 @@ transformers_base = dict(
     do_train=True,
     do_eval=True,
     dataloader_drop_last=True,  # keeps consistent batch size
-    num_train_epochs=3  # is overriden if max_steps is defined
+    num_train_epochs=3,  # is overridden if max_steps is defined
+    ddp_find_unused_parameters=False,  # Should not have any unused parameters
 )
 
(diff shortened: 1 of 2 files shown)


projects/transformers/experiments/base.py
```diff
@@ -46,7 +46,8 @@
     do_train=True,
     do_eval=True,
     dataloader_drop_last=True,  # keeps consistent batch size
-    num_train_epochs=3  # is overriden if max_steps is defined
+    num_train_epochs=3,  # is overridden if max_steps is defined
+    ddp_find_unused_parameters=False,  # Should not have any unused parameters
```
lucasosouza, line 50: That is interesting, I just looked into the documentation of torch.nn.parallel.DistributedDataParallel to understand what `find_unused_parameters` is doing. Did we observe run time improvements by adding this argument?

lucasosouza (approved):

projects/transformers/experiments/base.py
lscheinkman: Yes, It slowed the forward pass.

it is set to `False` by default in pytorch to avoid traversing the autograd graph during the forward pass.
See https://pytorch.org/docs/stable/notes/ddp.html#internal-design

PR #536, opened 2021-06-18
Dendritic MLPs with no dendrites

subutai:
Experiments for dendritic MLPs with no dendrites

Allow dendritic MLPs to have no segments.
Experiments testing continual learning with sparse non-dendritic networks.
Experiments testing a batch version of permuted MNIST on sparse non-dendritic networks.
Consolidated some experiment files and renamed them to be clearer.


projects/dendrites/permutedMNIST/experiments/batch_pmnist.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.current_epoch >= self.epochs
+
+    def run_task(self):
+        self.train_loader.sampler.set_active_tasks(range(self.num_tasks))
+        self.val_loader.sampler.set_active_tasks(range(self.num_tasks))
+        self.logger.info("Training task %d, epoch %d...",
+                         self.current_task, self.current_epoch)
+
+        # Just run one epoch
+        ret = self.run_epoch()
```
karangrewal, line 64: For flexibility, `run_epoch` should be called multiple times, since I find it hard to imagine a single pass through num_tasks * 60k images would be sufficient.

karangrewal: Also, the config specifies `epochs=3` below, so that's not being used.

karangrewal (approved):
I recommend removing "pmnist" from file names, since they're all in the "permutedMNIST" folder. This will keep file names cleaner and shorter. However, you can keep "batch_mnist" since that one is an exception and uses a slightly different dataset.

projects/dendrites/permutedMNIST/experiments/batch_pmnist.py
subutai: I believe it does run it through `epochs` times. `run_epoch()` increments `self.current_epoch`. `run_task()` is just a pass through here to run one epoch and is called repeatedly until `should_stop()`, redefined on line 54, is `True`. So essentially it runs through epochs the same way a supervised experiment does, but retains the rest of the code path of `DendriteContinualLearningExperiment`.

subutai:
> I recommend removing "pmnist" from file names, since they're all in the "permutedMNIST" folder. This will keep file names cleaner and shorter. However, you can keep "batch_mnist" since that one is an exception and uses a slightly different dataset.

Yeah, I was distinguishing `batch_pmnist` from `batch_mnist`. Should I just rename `batch_pmnist.py` to `batch.py`?

projects/dendrites/permutedMNIST/experiments/batch_pmnist.py
karangrewal: If `run_epoch` was called multiple times when you ran this, I'll  believe that. Then, the following line can be placed in `setup_experiment`, and doesn't need to be called repeatedly, right?

```
self.train_loader.sampler.set_active_tasks(range(self.num_tasks))
```

karangrewal:
> > I recommend removing "pmnist" from file names, since they're all in the "permutedMNIST" folder. This will keep file names cleaner and shorter. However, you can keep "batch_mnist" since that one is an exception and uses a slightly different dataset.
> 
> Yeah, I was distinguishing `batch_pmnist` from `batch_mnist`. Should I just rename `batch_pmnist.py` to `batch.py`?

Yes, I think you can rename:

`batch_pmnist.py` --> `batch.py`
`pmnist_no_dendrites.py` --> `no_dendrites.py`

PR #550, opened 2021-07-20
RES-2198: Add sparse deepspeed transformer layer configuration and mixin

lscheinkman:
Sparse version of https://www.deepspeed.ai/tutorials/transformer_kernel/ 
Got 10% speed up over the original HF implementation

### DIFF
--- a/projects/transformers/models/sparse_deepspeed_transformer.py
+++ b/projects/transformers/models/sparse_deepspeed_transformer.py
@@ -43,30 +43,41 @@ class SparseDeepSpeedTransformerLayer(SparseWeightsBase):
         assert isinstance(module, DeepSpeedTransformerLayer)
         super().__init__(module=module, sparsity=sparsity)
 
-        # Sparsify fused attention layer
+        # Sparsify fused attention layer, applying local sparsity to each QKV layer
         device = module.attn_qkvw.device
         out_features = module.attn_qkvw.shape[0] // 3
         in_features = module.attn_qkvw.shape[1]
         mask_qw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
         mask_kw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
         mask_vw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
-        zero_mask_attn_qkv = torch.cat((mask_qw, mask_kw, mask_vw), dim=0)
-        self.register_buffer("zero_mask_attn_qkv", zero_mask_attn_qkv)
+        zero_mask_attn_qkvw = torch.cat((mask_qw, mask_kw, mask_vw), dim=0)
 
         # Sparsify attention output layer
-        zero_mask_attn_out = _compute_zero_mask(
+        zero_mask_attn_ow = _compute_zero_mask(
             *module.attn_ow.shape, sparsity=self.sparsity, device=device)
-        self.register_buffer("zero_mask_attn_out", zero_mask_attn_out)
 
         # Sparsify intermediate layer
-        zero_mask_inter = _compute_zero_mask(
+        zero_mask_inter_w = _compute_zero_mask(
             *module.inter_w.shape, sparsity=self.sparsity, device=device)
-        self.register_buffer("zero_mask_inter", zero_mask_inter)
 
         # Sparsify output layer
-        zero_mask_output = _compute_zero_mask(
+        zero_mask_output_w = _compute_zero_mask(
             *module.output_w.shape, sparsity=self.sparsity, device=device)
-        self.register_buffer("zero_mask_output", zero_mask_output)
+
+        # Compute contiguous weight tensor and mask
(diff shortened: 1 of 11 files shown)


lucasosouza (approved):
My only concern is whether this new layer maintains compatibility with the dynamic sparsity algorithms, mainly RigL and GMP, I can't tell. I will tag @mvacaporale

lscheinkman:
It is definitely not compatible with RigL because it is a deepspeed specific layer. 
I will update the PR to make sure it stays compatible with GMP as long as the sparsity is kept global for the whole bert layer.

projects/transformers/trainer_mixins/deepspeed.py
```diff
@@ -0,0 +1,129 @@
...
+    # Store sparse mask and properties before deepspeed conversion
+    sparsity_per_layer = defaultdict(dict)
+    for i, bert_layer in enumerate(model.encoder.layer):
+        # Only supports huggingface BertLayer
+        assert isinstance(bert_layer, BertLayer)
+
+        # Check if model has sparse bert layers
+        query = bert_layer.attention.self.query
+        if isinstance(query, SparseWeights):
+            # Assume same sparsity for all QKV layers
```
mvacaporale, line 47: We have functionality to set a different sparsity per layer, though I've only used it in a a couple experiments. It could be good to specify this assumption in the docstring of the mixin.

mvacaporale: That said, you should definitely keep this assumption. Most likely, all of our experiments will abide by it.

projects/transformers/trainer_mixins/deepspeed.py
```diff
@@ -0,0 +1,129 @@
...
+    # SparseDeepSpeedTransformerLayer and restore the original zero mask and
+    # sparsity configuration
+    for i, sparse_config in sparsity_per_layer.items():
+        ds_layer = ds_model.encoder.layer[i]
+
+        attn_qkv_sparsity, attn_qkv_mask = sparse_config["attn_qkv"]
+        attn_out_sparsity, attn_out_mask = sparse_config["attn_out"]
+        inter_sparsity, inter_mask = sparse_config["inter"]
+        output_sparsity, output_mask = sparse_config["output"]
+
```
mvacaporale, line 97: You don't use the `attn_out_sparsity`, `inter_sparsity`, and `output_sparsity` it seems. Should you just not define them so it's clear that the sparsity of `attn_qkv_sparsity` will be used for the module?

projects/transformers/models/sparse_deepspeed_transformer.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Sparsify fused attention layer
+        device = module.attn_qkvw.device
+        out_features = module.attn_qkvw.shape[0] // 3
+        in_features = module.attn_qkvw.shape[1]
+        mask_qw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
+        mask_kw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
+        mask_vw = _compute_zero_mask(out_features, in_features, self.sparsity, device)
+        zero_mask_attn_qkv = torch.cat((mask_qw, mask_kw, mask_vw), dim=0)
+        self.register_buffer("zero_mask_attn_qkv", zero_mask_attn_qkv)
```
mvacaporale, line 54: Unfortunately, this won't be compatible with the GMP mixin. In particular, we need this module to be compatible with the `global_prune_by_abs_weight` function of `global_pruning.py`. It assumes each `SparseWeightsBase` class instance has an attribute `weight` and `zero_mask`.

One possibility is to add property methods for `weight` and `zero_mask` that concatenates all the weights into a flattened vector and all the masks into a flattened vector. Then we could add a setter method for `zero_mask` that takes a flattened mask and reshapes the subsections accordingly to set `zero_mask_attn_qkv`, `zero_mask_attn_out`, and so on.

mvacaporale (commented):
The PR looks great. However, the `SparseDeepSpeedTransformerLayer` won't be compatible with the GMP mixin. See comment below.

projects/transformers/tests/deepspeed_test.py
```diff
@@ -0,0 +1,125 @@
...
+                labels=labels[i],
+                token_type_ids=token_type_ids[i],
+            )
+            expected_loss = expected_outputs.loss
+
+            actual_loss.backward()
+            expected_loss.backward()
+
+            # Make sure model returns same results
+            self.assertTrue(torch.isclose(actual_loss, expected_loss, atol=1e-2))
```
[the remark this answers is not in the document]
lscheinkman: Done af99b49ef065becf013969475021e00831231a48

projects/transformers/trainer_mixins/deepspeed.py
lscheinkman: Done beb8eb7bcd8fb8a37c3e4fc1b111d4b250cd766f

PR #552, opened 2021-07-22
Configs for Small BERT Models

mvacaporale:
These are the configs used to run the 80, 85, 90, 2x, and 4x models


benja-matic (commented):
Looks good. Gonna go ahead and merge this so I can use the configs to kick off finetuning.

PR #553, opened 2021-07-23
Update Readme, cleanup at the end of runs, RES-2344, RES-2342, RES-2334

benja-matic:
- Removed table from readme, created a results directory with tables and documentation
- Stripped down results table to exclude old runs and runs that had bugs
- Added code to rm checkpoints and unnecessary directories at the end of training
- This (tentatively) solved ray issue by removing files taking up excess space
- Moved hyperparameter storage to yaml instead of pickle for transparency
- Removed ray.init() and ray.shutdown() in run.py, which were causing problems and providing no obvious benefit
- Cleaned up finetuning and trifecta configs
- Added steps to ensure nans don't get added to hyperparameters due to joining dataframes with different columns


mvacaporale (approved):
Looks good. Thanks for wrangling on the fine-tuning files and results. The one thing I'd caution is that the run functionality (such as `run_utils.py`) may be accumulating technical debt. There's a lot we're trying to accomplish between pre-training, fine-tuning, and hyper-parameter search for both of these. At some point in the future, we may benefit from slowing down and spending a couple days or so writing tests, refactoring, and distilling down the core functionality.

PR #556, opened 2021-08-13
Multiple Eval Sets MNLI

benja-matic:
- Created a mixin where trainer.evaluate() loops over a list of evaluation sets
- Refactored and fixed various checks, including new checks to ensure the above mixin is used when finetuning mnli
- Making a good default behavior for handling mnli required several adjustments, including new run_args and updates to task_hyperparams updates in run.py
- Added configs for numerous experiments
- Removed some configs that are no longer in use to avoid confusion, added labels
- Gathered hyperparameters for small berts
- Added hyperparam analysis code so that hps for one task serve as a proxy for another
- Updated the prediction and link predictions components of run_utils to handle multiple test sets


projects/transformers/callbacks/metrics.py
```diff
@@ -25,104 +25,103 @@
...
-        if is_mnli_mm:
-            self.mm_metrics = {ksplit(k): v for k, v in metrics.items()}
+        return do_remaining_updates
 
-        return is_mnli_mm
+    def on_train_begin(self, args, state, control, **kwargs):
+        """
+        Stop from accumulating results accross multiple runs
+        """
+        self.__init__()
```
lucasosouza, line 95: Is this the same thing as the original init, shouldn't it take arguments? Is there another way of reinitializing result tracking without calling init?

projects/transformers/trainer_mixins/multi_eval_sets.py
```diff
@@ -0,0 +1,85 @@
...
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+
+class MultiEvalSetsTrainerMixin:
```
lucasosouza, line 23: This looks useful, to have it generalized. But introducing a class that handles the case of multiple eval sets begs the question of whether there is any use case except MLNI

projects/transformers/run.py
```diff
@@ -441,16 +471,28 @@ def model_init():
...
+
+    success = False
+    try:
+        best_run = trainer.hyperparameter_search(**hp_search_kwargs)
+        logging.info(f"Best run: {best_run}")
+        success = True
+    except TuneError:
+        logging.info(f"One or more trials errored out")
+    finally:
+        logging.info("An unknown error occured during hp search.")
```
lucasosouza, line 485: This probably doesn't belong in the finally block, but in the except block - the finally block will execute regardless where it fails or not. So even if it succeeds you will get an error message saying an unknown error occurred.

Finally is only useful if you wanna make sure things like saving X or closing Y is always done regardless of whether the code executes or not.

projects/transformers/run.py
```diff
@@ -441,16 +471,28 @@ def model_init():
...
 
     # Delete all saved models and checkpoints to save space. All we need
     # are the params and scores. Currently this is a hack that is specific
     # to ray. May need to modify so that it serves the same function for
     # sigopt, once I integrate it.
     rm_prefixed_subdirs(training_args.output_dir, "run-")
 
+    if not success:
+        return {}
```
lucasosouza, line 494: Return is {} anyway - this might be better arranged with a `if success: code` and a single return at the end. The return here would imply if not success you would have a different output. 

Also, it is a bit weird that this function always returns an empty dict. Shouldn't it be returning the eval results instead?

projects/transformers/run.py
```diff
@@ -576,27 +614,31 @@ def run_finetuning_multiple_tasks(
...
         data_args.task_name = task_name
         training_args = deepcopy(base_training_args)
+        model_args = deepcopy(base_model_args)
         # For each task, save to a subfolder within run's root folder
         training_args.run_name = f"{base_training_args.run_name}_{task_name}"
         training_args.output_dir = os.path.join(
             base_training_args.output_dir, task_name
         )
 
+        model_arg_keys = ["trainer_class", "task_hyperparams_proxy"]
```
lucasosouza, line 628: Is that the case we have different trainer_class for each finetuning task? If not, and task_hyperparams_proxy is the only argument that changes, couldn't that also be part of training_args so we don't have to include model_args in the hp loop as well?

lucasosouza (changes requested):
Looks good. Left as request changes because there is an improper use of a finally block in run.py, which is an easy fix - otherwise the other comments are questions and optional suggestions. I was more confused about a few changes in run.py

projects/transformers/callbacks/metrics.py
```diff
@@ -25,104 +25,103 @@
...
         Example:
-            self.eval_metrics['acc'] -> [acc1, acc2, ..., accn]
+            self.eval_metrics['acc'] -> [acc_1, acc_2, ..., acc_n]
             self.steps = [eval_steps, eval_steps*2, ..., eval_steps*n]
 
-        This callback also keeps track of model sparsity, and breaks if
-        sparsity changes in absolute value by more than sparsity_tolerance.
-        If you are using a training approach that sparsifies the model, be sure
-        to set sparsity_tolerance to something like 1, so large changes in
-        sparsity are accepted.
```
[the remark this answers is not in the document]
benja-matic: This was replaced with properly formatted documentation. That comment is still in the docstring.

projects/transformers/callbacks/metrics.py
benja-matic: Good catch. That absolutely should have arguments. RE not calling init, the alternative would be to manually reset variables like self.step_counter and self.eval_metrics. But that's all the init does anyway, so I chose to call it.

projects/transformers/trainer_mixins/multi_eval_sets.py
benja-matic: For now the only use case is MNLI. However, I think a mixin like this could be useful for other projects or different types of experiments (multi task learning, meta learning, continual learning). If you have a family of tasks, it could be informative to see how performance on each one changes over time, while training on a source dataset.

projects/transformers/run.py
benja-matic: Whoops, good catch here as well.

benja-matic: In the updated code, the finally block is used to call the code to remove unnecessary checkpoints. That does need to run regardless of rather or not hp search completes.

projects/transformers/run.py
benja-matic: It is a little funky. It returns {} because run_finetuning_multiple_tasks sets up aloop over tasks / runs, and calls run_finetuning_single_task or run_finetuning_single_task_with_hp_search inside it. The former returns task results. The latter is just used to find good hyperparams and is not intended (currently) for producing finetuning results, so it returns something empty that is the same type. The reason for returning early is that if hp search didn't complete, best_run won't be defined and the rest of the code will break.

This raises two changes: 1) I can just define best_run with a null value so the "if not success" statement can be removed, or simply delete the code to save best_run to a text file since this is used for redundancy, and is not the primary way of analyzing hp search results. 2) The thing returned in run_finetuning_single_task, eval_results, isn't even defined unless training_args.do_evaluate. I haven't seen the code break because of this, since do_eval is the default and there are checks / warnings elsewhere, but it needs fixed.

projects/transformers/run.py
benja-matic: The same trainer class is used for all tasks except mnli, which requires a class that uses the multi_eval_sets mixin. Model args is also needed because ("hp_" in hp_key) is checking to update things like hp_num_trials and hp_compute_objective. I admit this is a little messy, but I'm prioritizing gathering results atm, so I'm gonna roll with it.

PR #557, opened 2021-08-17
First PR - updating nupic.research README

abhi-iyer:
Added reference to nupic.embodied repository at the end of README


README.md
```diff
@@ -45,7 +45,13 @@ Otherwise you need can install using setup.py like any python project. Since the
 You can test your installation by running the test script from the repository root:
 
     python setup.py test
-  
+
+Active Projects
+=======
+
+Some of our active research code and experiments are contained in the following repositories:
```
lucasosouza, line 52: Looks good. I would just change "are contained in" to "can also be found in", so it is more clear that we have code in other repositories as well.

lucasosouza (changes requested):

lucasosouza:
@abhi-iyer One other thing I forgot to mention, we always push from a branch, never from `master` - so you should create a local branch in your repository, and then create that PR from that branch into numenta/master

PR #559, opened 2021-08-19
Greedy Infomax Cleanup and Refactor

alexcuozzo-1:
https://ntaresearch.atlassian.net/browse/RES-2327


nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2021, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
```
karangrewal, line 4: Files should be named using pothole case (e.g., `bilinear_info.py`) and not camel case (e.g., `BilinearInfo.py`). In contrast, python classes are named using camel case. Please change file names to pothole case where necessary:

- `BilinearInfo.py`
- `ClassificationModel.py`
- `FullModel.py`
- `ResNetEncoder.py`

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
...
+    This model computes:
+
+        log f(z_{t+k}, z_{t}) = z_{t+k} W_k z_{t}
+
+    for each prediction step k along with a specified number of random contrastive
+    samples.
+
+    After computing this value for all necessary positive and contrastive samples,
+    the positive examples are at index 0 and the rest are indices 1:n, where n is
+    the number of contrastive samples per positive sample. Therefore, the the loss
```
karangrewal, line 54: Typo.

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
...
+                device=ztwk_shuf.device,
+            )
+            rand_index = rand_index.repeat(1, ztwk_shuf.shape[1])
+            ztwk_shuf = torch.gather(
+                ztwk_shuf, dim=0, index=rand_index, out=None
+            )  # y * x * b * n, c
+            ztwk_shuf = ztwk_shuf.view(
+                ztwk.shape[0],
+                ztwk.shape[1],
+                ztwk.shape[2],
```
karangrewal, line 123: Style comment: this can be reduced to fewer lines, instead of having each argument on a separate line.

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
...
+        out_channels,
+        negative_samples=16,
+        k_predictions=5,
+        sparse_weights_class=SparseWeights2d,
+        sparsity=0.1,
+    ):
+        super(SparseBilinearInfo, self).__init__(
+            in_channels, out_channels, negative_samples, k_predictions
+        )
+        if sparsity and sparsity > 0.3:
```
karangrewal, line 172: By default, `sparsity=0.1`, so why do you need the `if sparsity` clause here? I assume a user wouldn't initialize an instance of `SparseBilinearInfo` if they don't want sparsity, so a check shouldn't be necessary.

nupic/research/frameworks/greedy_infomax/models/FullModel.py
```diff
@@ -0,0 +1,911 @@
...
+    ):
+        super(SparseFullVisionModel, self).__init__(
+            negative_samples=negative_samples,
+            k_predictions=k_predictions,
+            resnet_50=resnet_50,
+            grayscale=grayscale,
+            patch_size=patch_size,
+            overlap=overlap,
+        )
+        if sparsity is None:
```
karangrewal, line 204: Same as above; a user can just initiate a `FullVisionModel` instead of `SparseFullVisionModel` if there's no sparsity, correct? Either assume `sparsity` is a value in (0, 1) or raise an error if it's 0. Thoughts?

nupic/research/frameworks/greedy_infomax/models/FullModel.py
```diff
@@ -0,0 +1,911 @@
...
+    ):
+        super(SparseSmallVisionModel, self).__init__(
+            negative_samples=negative_samples,
+            k_predictions=k_predictions,
+            resnet_50=resnet_50,
+            grayscale=grayscale,
+            patch_size=patch_size,
+            overlap=overlap,
+        )
+        if sparsity is None:
```
karangrewal, line 534: Same `sparsity` concern.

nupic/research/frameworks/greedy_infomax/models/FullModel.py
```diff
@@ -0,0 +1,911 @@
...
+        super(SuperGreedySparseSmallVisionModel, self).__init__(
+            negative_samples=negative_samples,
+            k_predictions=k_predictions,
+            resnet_50=resnet_50,
+            grayscale=grayscale,
+            patch_size=patch_size,
+            overlap=overlap,
+        )
+        if sparsity is None:
+            # reverts to dense weights
```
karangrewal, line 682: Same `sparsity` concern.

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
...
+            for i in range(len(self.W_k)):
+                self.W_k[i] = sparse_weights_class(self.W_k[i], sparsity=sparsity)
+
+
+class VDropSparseBilinearInfo(BilinearInfo):
+    def __init__(
+        self,
+        in_channels,
```
karangrewal, line 180: Should this really be a separate class, or should `BilinearInfo` / `SparseBilinearInfo` just have a regularization option that allows for VDrop?

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
```diff
@@ -0,0 +1,198 @@
...
+            log_f_list.append(log_fk)
+            true_f_list.append(true_fk)
+
+        return log_f_list, true_f_list
+
+
+class SparseBilinearInfo(BilinearInfo):
+    def __init__(
+        self,
+        in_channels,
```
karangrewal, line 162: Please add documentation for this class, especially since since the `__init__` here has has args like `sparsity` that aren't used in the parent class.

nupic/research/frameworks/greedy_infomax/models/ClassificationModel.py
```diff
@@ -0,0 +1,71 @@
...
+        self.model.add_module(
+            "layer1", nn.Linear(self.in_channels, num_classes, bias=True)
+        )
+
+    def forward(self, x):
+        # detach x just in case it's still connected to active parts of the
+        # computation graph
+        x = x.detach()
+        x = self.avg_pool(x).squeeze()
+        x = self.model(x).squeeze()
```
karangrewal, line 45: Calling `squeeze` when `x` has batch size 1 will result in removing a dimension (since the batch dimension would have only 1 element). Are you sure you need the `squeeze` method?

nupic/research/frameworks/greedy_infomax/models/ClassificationModel.py
```diff
@@ -0,0 +1,71 @@
...
+# https://arxiv.org/abs/1905.11786
+# ----------------------------------------------------------------------
+
+import torch.nn as nn
+
+
+class ClassificationModel(nn.Module):
+    def __init__(self, in_channels=256, num_classes=10, hidden_nodes=0):
+        super().__init__()
+        self.in_channels = in_channels
```
karangrewal, line 33: Please add documentation for `ClassificationModel` and `MultipleClassificationModel`. Also, I think `Classifer` would be a better name for this class - thoughts?

nupic/research/frameworks/greedy_infomax/models/ResNetEncoder.py
```diff
@@ -0,0 +1,767 @@
...
+        out = self.nonlinearity3(out)
+        out = self.conv3(out)
+        out += shortcut
+        return out
+
+
+class ResNetEncoder(nn.Module):
+    """
+    The main subcomponent of FullVisionModel. This encoder also implements both
+    .forward() and .encode() to support different outputs for unsupervised and
```
karangrewal, line 368: It turns out there's already some implementation of ResNets (or maybe just parts of them) in nupic.research. Check out the [pytorch folder](https://github.com/numenta/nupic.research/tree/master/nupic/research/frameworks/pytorch/models). Is there anything there you can re-use for any of your model implementation? We should avoid duplicate code wherever possible

nupic/research/frameworks/greedy_infomax/utils/data_utils.py
```diff
@@ -0,0 +1,73 @@
...
+from copy import deepcopy
+
+from torchvision import transforms
+
+
+# get transforms for the dataset
+def get_transforms(val=False, aug=None):
+    trans = []
+
+    if aug["randcrop"]:
```
karangrewal, line 37: Documentation!

nupic/research/frameworks/greedy_infomax/utils/data_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    trans.append(transforms.Normalize(mean=aug["bw_mean"], std=aug["bw_std"]))
+
+    trans = transforms.Compose(trans)
+    return trans
+
+
+# labeled train set: mean [0.4469, 0.4400, 0.4069], std [0.2603, 0.2566, 0.2713]
+aug = {"randcrop": 64, "flip": True, "bw_mean": [0.4120], "bw_std": [0.2570]}
+transform_unsupervised = get_transforms(val=False, aug=aug)
+transform_validation = transform_supervised = get_transforms(val=True, aug=aug)
```
karangrewal, line 57: I believe the code here and everything below should be in the projects folder, since it's experiment-specific. It doesn't make sense to have this in frameworks.

nupic/research/frameworks/greedy_infomax/utils/loss_utils.py
```diff
@@ -0,0 +1,91 @@
...
+# ----------------------------------------------------------------------
+
+import torch
+import torch.nn.functional as F
+
+
+def multiple_cross_entropy(data_lists, targets, reduction="mean"):
+    log_f_module_list, true_f_module_list = data_lists
+    device = log_f_module_list[0][0].device
+    total_loss = torch.tensor(0.0, requires_grad=True, device=device)
```
karangrewal, line 34: I don't think 'multiple cross entropy' is widely-used term, so please add documentation for this function.

nupic/research/frameworks/greedy_infomax/utils/model_utils.py
```diff
@@ -0,0 +1,335 @@
...
+    x = x.reshape(
+        x.shape[0] * x.shape[1] * x.shape[2], x.shape[3], x.shape[4], x.shape[5]
+    )
+    return x, n_patches_x, n_patches_y
+
+
+def train_model_gim(
+    model,
+    loader,
+    optimizer,
```
karangrewal, line 77: Style comment: to be consistent with other train/eval functions, consider changing the name here to `train_gim_model` (and likewise for the eval function below). This would be consistent with how we defined [train & eval functions for dendrites](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/dendrites/model_utils.py#L33-L50).

nupic/research/frameworks/greedy_infomax/utils/model_utils.py
```diff
@@ -0,0 +1,335 @@
...
+    :rtype: float
+    """
+    model.train()
+    # Use asynchronous GPU copies when the memory is pinned
+    # See https://pytorch.org/docs/master/notes/cuda.html
+    async_gpu = loader.pin_memory
+    if progress_bar is not None:
+        loader = tqdm(loader, **progress_bar)
+        # update progress bar total based on batches_in_epoch
+        if batches_in_epoch < len(loader):
```
karangrewal, line 137: Here, you have defined your own custom train function. However, if you don't need some of these utilities (e.g., tqdm), you can delete the code and just say "Unused" for the corresponding parameter in the docstring.

See how we created a new and less-convoluted train function which is specific to dendrites, [here](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/dendrites/model_utils.py#L33-L50). If you delete a lot of these things that you know won't be used, that will seriously clean up the code here.

karangrewal: The same goes for your eval function.

nupic/research/frameworks/vernon/mixins/regularize_loss.py
```diff
@@ -80,7 +80,7 @@ def pre_batch(self, batch_idx, **kwargs):
 
     def complexity_loss(self, model):
         c_loss = super().complexity_loss(model)
-        assert model is self.model
+        # assert model is self.model
```
karangrewal, line 83: Two things:

1. This is the loss function and will be called regularly, so having an `assert` statement will slow down the overall runtime; you should consider removing it.
2. If there's code that you don't need, delete it instead of commenting it out.

karangrewal: Ah, you commented out the `assert` statement -- my bad. In that case, only #2 above applies.

nupic/research/frameworks/vernon/mixins/sync_batchnorm.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+import torch.nn as nn
+
+
+class SyncBatchNorm:
+    def create_model(self, config, device):
+        model = super().create_model(config, device)
+        use_synch_batchnorm = config.get("use_synch_batchnorm", True)
```
karangrewal, line 27: Please add documentation for this mixin, so that it can also easily be later on by others. Also, since you're assuming the config contains certain args like `use_synch_batchnorm`, please add an example config in your docstring like is [here](https://github.com/numenta/nupic.research/blob/master/nupic/research/frameworks/vernon/mixins/dendrite_metrics.py#L97-L107) for this other mixin.

karangrewal: I assume 'use_synch_batchnorm' is a typo?

nupic/research/frameworks/vernon/mixins/prune_low_snr.py
```diff
@@ -100,7 +100,10 @@ def pre_epoch(self):
 
     def run_epoch(self):
         results = super().run_epoch()
-        vdrop_data = self.model.module.vdrop_data
+        model = self.model
+        if hasattr(model, "module"):
+            model = model.module
```
karangrewal, line 105: Are you sure these changes are compatible with how this mixin is used elsewhere? It seems like you've made it custom to fit your case: the line you removed (old 103) would be equivalent of `self.model.module.module.vdrop_data` under your new implementation, which looks like an inconsistency to me.

nupic/research/frameworks/vernon/run.py
```diff
@@ -133,11 +133,12 @@ def head(_, config, world_size, logger, on_checkpoint, queue, results):
             epoch_result = worker_results[epoch]
             if len(epoch_result) == world_size:
                 epoch_result = experiment_class.aggregate_results(epoch_result)
-                epoch_result.update(epoch=epoch,
-                                    timestamp=time.time(),
-                                    pid=os.getpid(),
-                                    hostname=os.uname()[1],
```
karangrewal, line 139: The multi-lining style in this file is consistent with the rest of nupic.research, so you don't need whatever style tool you used (possible black). You can revert all multi-lining changes in this file to the way they originally were.

projects/greedy_infomax/experiments/default_base.py
```diff
@@ -0,0 +1,256 @@
...
+    # Dataset
+    dataset_class=STL10,
+    dataset_args=STL10_DATASET_ARGS,
+    #               STL10:
+    # 500 training images (10 pre-defined folds)
+    # 8000 test images (800 test images per class)
+    # 100,000 unlabeled images
+    # num_unsupervised_samples=1000,
+    # num_supervised_samples=500,
+    # num_validation_samples=32,
```
karangrewal, line 111: In this config, you have a lot of lines of code which are commented out. You should delete these if they're not used.

karangrewal (changes requested):

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
alexcuozzo-1: In other models, VDrop versions are explicitly created as classes. In this case, it's necessary to introduce the central_data parameter which is specific to VDrop.

nupic/research/frameworks/greedy_infomax/models/ClassificationModel.py
alexcuozzo-1: I do need the squeeze here, but have added extra code to ensure that the batch dimension is preserved.

nupic/research/frameworks/greedy_infomax/models/ClassificationModel.py
alexcuozzo-1: That works and I can change it. The reason for being called ClassificationModel (and all the other models' names) is that they were taken from the original codebase. Considering we are no longer even referencing the original codebase, I am happy to change it.

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
alexcuozzo-1: Yes, changed.

nupic/research/frameworks/greedy_infomax/models/FullModel.py
alexcuozzo-1: In this case, sparsity is actually a hierarchy of dicts (in order to specify the sparsity of every layer in the network), which is why it defaults to None instead of 0.

nupic/research/frameworks/greedy_infomax/models/FullModel.py
alexcuozzo-1: see above

nupic/research/frameworks/greedy_infomax/models/FullModel.py
alexcuozzo-1: see above

nupic/research/frameworks/greedy_infomax/models/ResNetEncoder.py
alexcuozzo-1: I thought about this when I was implementing the codebase. Yes, it is possible to go back and use parts of the current codebase but I decided against it because:
1) Code reproducibility. Maybe we implemented our ResNet in a slightly different way than the GIM paper, which would not lead to comparable results
2) Instead of using a BasicBlock, the paper uses a modified version of the block with no batch norm, which we do not have in resnets.py
3) The ResNet class includes both the NoBatchnormBlock and also the BilinearInfo block. In order to use the BilinearInfo like the paper, I would have to create a new class anyway

It's still possible to reuse a bit of the code, but it will take a bit of time to make sure that the implementations are comparable again. Plus, we have now moved over to the BlockModel. I'm trying to walk the line between codebase maintenance and moving forward with the goals of the project. Open to suggestions.

nupic/research/frameworks/greedy_infomax/utils/data_utils.py
alexcuozzo-1: Done

nupic/research/frameworks/greedy_infomax/utils/data_utils.py
alexcuozzo-1: Done

nupic/research/frameworks/greedy_infomax/models/BilinearInfo.py
alexcuozzo-1: Done

nupic/research/frameworks/greedy_infomax/utils/loss_utils.py
alexcuozzo-1: Done

nupic/research/frameworks/greedy_infomax/utils/model_utils.py
alexcuozzo-1: Done

nupic/research/frameworks/greedy_infomax/utils/model_utils.py
alexcuozzo-1: Done

nupic/research/frameworks/vernon/mixins/sync_batchnorm.py
alexcuozzo-1: Done

nupic/research/frameworks/vernon/mixins/prune_low_snr.py
alexcuozzo-1: Good catch, fixed

nupic/research/frameworks/vernon/run.py
alexcuozzo-1: Done

projects/greedy_infomax/experiments/default_base.py
alexcuozzo-1: Removed commented lines

karangrewal (approved):
Please change the license year in `projects/greedy_infomax/run.py` to 2021 (from 2020) and you're all set. (Yes, I know it's an OCD comment).

PR #566, opened 2021-09-13
Cross hyperparameters search

jeremyforest:
config file for generating the cross hyperparameter search (data showed on dendrite meeting at the end of August)


karangrewal (changes requested):
I have 2 pieces of feedback:

1. I think it would make more sense in the config if you give the optimal parameter configuration instead of code that runs a hyperparameter search. This way, someone else can use your config to run an experiment with optimal choice of hyperparameters. E.g., you would have `num_segments=10` instead of `num_segments=tune.grid_search([5, 10, 50, 100])`. However, if these optimal hyparameter configurations are in some other config and this is just code to do the hyperparameter search, that's fine too.

2. Your Makefile doesn't belong in the repository, please remove that from your commit.

jeremyforest:
1. yes this is just the code for the search so it should stay as-is
2. makefile allows to run all the scripts generating the figure in one go. 
3. adding more guidlines in the readme file.

jeremyforest:
@karangrewal I think  this is good to merge. let me know if you have more to add.

karangrewal (approved):

PR #568, opened 2021-09-20
RES-2382: implemente squad v1

benja-matic:
- Question Answering trainer mixin
- code to register question answering model
- configs for dense and sparse squad runs, mainly for testing
- code to load, preprocess, postprocess squad
- skeleton code for beam search and squad v2


projects/transformers/callbacks/metrics.py
```diff
@@ -58,6 +58,8 @@ def __init__(self, n_eval_sets=1, sparsity_tolerance=0.01):
     def on_evaluate(self, args, state, control, metrics, **kwargs):
         """Update eval metrics and possibly step counter, sparsity, and lr"""
 
+        print(f"In Metrics Callback. Here are the metrics: {metrics}")
```
lucasosouza, line 61: Can we change this to the logging, so it can't be more easily controlled by the global verbosity level

projects/transformers/experiments/squad.py
```diff
@@ -0,0 +1,126 @@
...
+    pass
+
+
+debug_bert_squad_base = deepcopy(finetuning_bert_100k_glue_get_info)
+debug_bert_squad_base.update(
+    # Model Args
+    model_name_or_path="bert-base-cased",
+    finetuning=True,
+    task_names=None,
+    task_name="squad",
```
lucasosouza, line 45: There are a lot of hyperparameters are related. If task_name is squad, then dataset_name has to be squad and trainer_class has to be QuestionAnsweringTrainer. It is easy to incur into bugs if we forget to set some of these

We can add some code in run.py to check, and either throw an Exception, or make the changes and throw a Warning. If they were part of the same dataclass parser, another possible solution is to do that check in post_init in the dataclass. 

None of this is required, it is not a big deal, but it would help in avoiding bugs.

projects/transformers/finetuning_constants.py
```diff
@@ -46,7 +47,8 @@
     "qnli": "QNLI",
     "rte": "RTE",
     "wnli": "WNLI",
-    "": "AX"  # diagnostic set, not yet implemented
+    "": "AX",  # diagnostic set, not yet implemented
```
lucasosouza, line 50: what is AX?

projects/transformers/finetuning_constants.py
```diff
@@ -57,6 +59,7 @@
     "qnli": ["accuracy"],
     "qqp": ["accuracy", "f1"],
     "rte": ["accuracy"],
+    "squad": ["exact_match", "f1"],
```
lucasosouza, line 62: What sort of metric is exact match, is that a specific way of measuring performance?

projects/transformers/requirements.txt
```diff
@@ -28,4 +28,4 @@ wandb>=0.10.27,<0.11.0 # this can be updated in a future version of transformers
 cloudpickle>=1.6.0
 pickle5>=0.0.11
 ray==1.4.0  # required to use Ray Tune
-transformers==4.8.1  # 4.8.1 required to use Ray Tune
+transformers==4.8.1
```
lucasosouza, line 31: I would leave that comment there, it would be easy to forget why we set the version to be specifically 4.8.1 (and the information would be useful for a new employee dealing with this codebase).

projects/transformers/run.py
```diff
@@ -583,6 +633,135 @@ def run_finetuning_single_task(
...
+def run_finetuning_squad(
+    model_args, data_args, training_args, last_checkpoint=None, run_idx=None,
+):
+    """On a single task train, evaluate, and save results"""
+
+    # Make sure dataset name, task name, and version_2_with_negative
+    # match before loading the dataset
+    data_args = check_squad_version(data_args)
+
+    data_init = init_dataset_for_squad(model_args,
```
lucasosouza, line 645: Another way of doing this:
```
(tokenizer, data_collator, train_dataset, eval_dataset, ...) = init_dataset_for_squad(model_args,
```

Anything you keep in parenthesis you can break down into more than one line. It is ultimately a style choice, up to you. Personally I think the one above looks cleaner and more in line with the rest of this particular codebase (the latter being a more important factor).

projects/transformers/run.py
```diff
@@ -643,21 +824,32 @@ def run_finetuning_multiple_tasks(
...
 
             if model_args.hp_num_trials > 1:
-                eval_results = run_finetuning_single_task_with_hp_search(
-                    model_args,
-                    data_args,
-                    training_args,
-                    last_checkpoint=last_checkpoint
-                )
+                if data_args.task_name == "squad":
+                    print("Hp search for squad is not yet implemented")
```
lucasosouza, line 828: Would it be hard to extend the existing hp search method for squad? When I initially coded I planned to make it generic enough to handle hp search for any finetuning task - if that is not the case, then it is a fail of the initial design, and the todo should be to fix that hp search instead of making a new one.

projects/transformers/run_args.py
```diff
@@ -321,6 +321,69 @@ class DataTrainingArguments:
...
+    null_score_diff_threshold: float = field(
+        default=0.0,
+        metadata={
+            "help": "The threshold used to select the null answer: if the best answer "
+                    "has a score that is less than the score of the null answer minus "
+                    "this threshold, the null answer is selected for this example. "
+                    "Only useful when `version_2_with_negative=True."
+        },
+    )
+    doc_stride: int = field(
```
lucasosouza, line 358: Adding the `squad_` prefix to these new squad arguments would help in separating namespaces.

projects/transformers/run_utils.py
```diff
@@ -459,6 +484,286 @@ def preprocess_function(examples):
...
+        tokenized_examples = tokenizer(
+            examples[key1], examples[key2], **tokenizer_kwargs)
+
+        return tokenized_examples
+
+    if split == "train":
+
+        def preprocess_function(examples):
+
+            examples[question_column_name] = [q.lstrip() for q in examples[question_column_name]]  # noqa: E501
```
lucasosouza, line 515: Although I strongly disagree with the 80 chars per line requirement as I believe it incentivizes poor variable naming (a heated discussion in our group), we should avoid leaving tons of flake 8 exceptions in the code. For example an easy fix for this would be (if it fits the line char requirements)

```
examples[question_column_name] = [
    q.lstrip() for q in examples[question_column_name]
]
```

projects/transformers/run_utils.py
```diff
@@ -459,6 +484,286 @@ def preprocess_function(examples):
...
+            tokenized_examples = pre_pre_process(examples)
+            sample_mapping = tokenized_examples.pop("overflow_to_sample_mapping")
+            offset_mapping = tokenized_examples.pop("offset_mapping")
+
+            # Let's label those examples!
```
lucasosouza, line 520: If there is a big chunk of code here we got from HF and don't really understand neither assessed it, we should instead add a comment stating that. That would be useful for a future developer to know not to change this piece of code without further understanding it.

projects/transformers/run_utils.py
```diff
@@ -459,6 +484,286 @@ def preprocess_function(examples):
...
+        load_from_cache_file=not data_args.overwrite_cache,
+    )
+
+    if getattr(data_args, max_samples_key) is not None:
+        dataset = dataset.select(range(getattr(data_args, max_samples_key)))
+
+    return examples, dataset
+
+
+def get_squad_tokenizer_kwargs(data_args, pad_on_right, max_seq_length):
```
lucasosouza, line 675: Most of these functions are missing docstrings, would be useful to add at least a one-liner

projects/transformers/run_utils.py
```diff
@@ -921,7 +1271,7 @@ def check_best_metric(training_args, task_name, metric=None):
 
     print("metric configuration after checks: "
           f"{metric}, {greater_is_better}")
-    training_args.greater_is_beter = greater_is_better
+    training_args.greater_is_better = greater_is_better
```
lucasosouza, line 1274: That is a weird argument - shouldn't this just be a function of the type of metric used?

lucasosouza (approved):
Awesome! There is lots of cleaning to do in this code, to be honest. There is a specific function for all things squad related (config, dataset, model, preprocessing, etc...). We did that as an initial step, but ideally the second step would be to refactor it by finding commonalities and evaluate if we actually needed specific functions for squad. That would make it easier to maintain - if we replicate the functions, any small change required would have to be done in all functions, and that increases the risk of some of these legacy code becoming deprecated. It is also missing documentation, like docstrings, or clarifying comments.

That said, if it works and serves our purpose, it doesn't feel like a necessity to do that cleaning now. So up to you. 

I left questions for things I didn't understand and some minor comments on other things.

projects/transformers/finetuning_constants.py
benja-matic: A diagnostic dataset we haven't used so far that is included in the GLUE datasets, but not factored into the the GLUE score. Last row of this table: https://gluebenchmark.com/tasks.

projects/transformers/run_utils.py
benja-matic: Pretty much, yeah. In HF it is an argument. In our code, it's an argument and these checks are supposed to ensure the correct configuration. Ideally I agree it isn't an argument at all but simply something determined on the fly based on the metric.

projects/transformers/run.py
benja-matic: I originally had it the way you described, but then I couldn't get flake8 to accept the indentation so finally I just did the brute force approach.

projects/transformers/run.py
benja-matic: It's not a huge leap, I just didn't have time for it yet so I put this in to clarify that I haven't gotten to that yet. The main lift is the way squad datasets and models are initialized. First, squad datasets are initialized with separate code currently. I know you commented above that you don't want that to be the case, but it takes a lot of code to preprocess squad, so either we can call the preprocessing code separately in the hp search function, or we can make another big lift to merge squad code into finetuning code.  Second, as you may recall, we have to define a special model init and pass that to the trainer.  This is a small modification.

projects/transformers/callbacks/metrics.py
benja-matic: Forgot to remove that. This was for debugging.

PR #578, opened 2021-10-05
Infer prototype while training

karangrewal:
The `CentroidContext` mixin has been augmented so that now it can do either of the following during training:

1. create prototype contexts via task labels while training (just as before)
2. use a clustering technique to create prototypes in an online way while training (new!)

There is also a corresponding config file for experiments that infer prototype while training. Two additional things to note:

- The experiment/mixin attribute `contexts` is now a List containing a single Tensor (instead of being just a single Tensor) since lists are mutable, and we require `contexts` to be mutable since it is updated regularly by other functions,
- There isn't any formal documentation of the exact clustering procedure in the mixin itself, and we can redirect to an ArXiv paper that contains this info (once it becomes available).

### DIFF
--- a/nupic/research/frameworks/vernon/mixins/centroid_context.py
+++ b/nupic/research/frameworks/vernon/mixins/centroid_context.py
@@ -21,21 +21,44 @@
 
 import abc
 
+import numpy as np
 import torch
+from scipy.stats import f
 
 from nupic.research.frameworks.dendrites import evaluate_dendrite_model
 
 __all__ = [
     "CentroidContext",
+    "cluster_centroid_while_training",
     "compute_centroid",
     "infer_centroid",
+    "provide_centroid_while_training",
+    "should_cluster",
+    "two_sample_hotelling_statistic"
 ]
 
 
 class CentroidContext(metaclass=abc.ABCMeta):
     """
     When training a dendritic network, use the centroid method for computing context
-    vectors (that dendrites receive as input) for both training and inference.
+    vectors (that dendrites receive as input) at either 1) just inference, or 2) both
+    training and inference. If computing cetroids while training, a clustering
+    algorithm is used which performs statistical tests betweens an existing centroid's
+    exemplars and the new batch of data to determine whether said centroid should
+    represent said batch.
+
+    The dict `centroid_context_args` in the experiment config specifies whether the
+    context vector should be inferred during training. The example below gives default
+    values.
+
+    Example config:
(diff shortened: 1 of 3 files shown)


nupic/research/frameworks/vernon/mixins/centroid_context.py
```diff
@@ -109,16 +172,174 @@ def _train_centroid(data):
...
+                     context vector
+    :param n_samples_per_context: List of ints where entry i gives the number of
+                                  samples used to compute the ith context (i.e.,
+                                  `contexts[0][i]`)
+    :param subindices: List/Tensor/Array that can index contexts to select subindices;
+                       optional
+    """
+
+    def _cluster(data):
+        nonlocal contexts
```
subutai, line 195: Do you need this? I'm not familiar with `nonlocal`. Can you just do a closure instead?

subutai (changes requested):

nupic/research/frameworks/vernon/mixins/centroid_context.py
```diff
@@ -45,22 +68,59 @@ def setup_experiment(self, config):
...
+
+            # `contexts` is a list containing a single tensor since we require a
+            # mutable object
+            self.contexts = [torch.zeros((0, dim_context))]
+            self.n_samples_per_context = []
+
+            # Subsample features from input to form the context; we do this
+            # dimensionality reduction because in order to transform a T-squared
+            # variable to an F-distributed variable, the dimensionality of the context
+            # must be strictly less than the total number of data samples
```
jeremyforest, line 91: could that also explain why you don't necessarily have the 'perfect' number of clusters ?

projects/dendrites/permutedMNIST/experiments/cluster.py
```diff
@@ -0,0 +1,160 @@
...
+        dendrite_weight_sparsity=0.0,
+        weight_sparsity=0.5,
+        context_percent_on=0.05,
+    ),
+
+    centroid_context_args=dict(infer_while_training=True),
+
+    batch_size=256,
+    val_batch_size=512,
+    epochs=5,
```
jeremyforest, line 84: we used to do 3 epochs before no ?

jeremyforest: ah ok I see below it's updated as a function of cluster size. Is it for optimization ?

PR #614, opened 2021-11-19
Add framework package integration tests based on dendrites projects: RES-2528

chris-redfield:
this solves RES-2528


tests/integration/frameworks/dendrites/dendrite_integration_tests.py
```diff
@@ -0,0 +1,142 @@
...
+)
+
+
+class DendriteIntegrationTest(unittest.TestCase):
+    """
+    This is a test class for testing Dendrite integration
+    with Vernon and Continual Learning
+    """
+
+    def test_base_experiment(self):
```
lscheinkman, line 121: This test is specific to "SynapticIntelligence" mixin, the test name should reflect that. 
Maybe `test_si_mixin` or something like that

lscheinkman (approved):
👍 
One small comment on the name of the test

PR #620, opened 2021-12-07
Added unit tests for MultiCycleLR mixin

vkakerbeck:
Tests whether
1. The initial learning rate is the one specified in each cycles' config
2. The learning rate never exceeds the max lr or goes below the min lr specified for the current epoch
on various different CycleLR configurations.

### DIFF
--- a/tests/unit/frameworks/vernon/multi_cycle_lr_test.py
+++ b/tests/unit/frameworks/vernon/multi_cycle_lr_test.py
@@ -22,7 +22,6 @@
 import unittest
 
 import numpy as np
-from nupic.research.frameworks.vernon.mixins import multi_cycle_lr
 import torch
 from torchvision.datasets import FakeData
 from torchvision.transforms import ToTensor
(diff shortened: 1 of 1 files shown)


tests/unit/frameworks/vernon/multi_cycle_lr_test.py
```diff
@@ -0,0 +1,168 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2020, Numenta, Inc.  Unless you have an agreement
```
lscheinkman, line 3: 2021

tests/unit/frameworks/vernon/multi_cycle_lr_test.py
```diff
@@ -0,0 +1,168 @@
...
+    ),
+)
+
+
+class MultiCycleLRTest(unittest.TestCase):
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
lscheinkman, line 118: You can just delete these empty methods

lscheinkman (approved):
👍 LGTM

tests/unit/frameworks/vernon/multi_cycle_lr_test.py
```diff
@@ -0,0 +1,166 @@
...
+                # If new cycle is started, check that initial lr is
+                # max_lr/div_factor
+                initial_lr = (
+                    current_lr_conf["max_lr"] / current_lr_conf["div_factor"]
+                )
+                self.assertAlmostEqual(lr[0], initial_lr)
+
+            # Check that lr of this epoch stays below max lr according to
+            # multiCycleLR schedule.
+            assert np.max(lr) <= current_lr_conf["max_lr"]
```
subutai, line 158: Unless I'm missing something, we shouldn't use assert's here - should use some form of self.assert...()

subutai (changes requested):

tests/unit/frameworks/vernon/multi_cycle_lr_test.py
lscheinkman: Good point. 
Using [self.assert...()](https://docs.python.org/3/library/unittest.html#assert-methods) instead of plain `assert` allow the runner to accumulate the results

vkakerbeck: Alright, I updated this: https://github.com/numenta/nupic.research/pull/620/commits/5e1d69f99dd42247df9f4b96db254d7c30f7ba22

subutai (approved):

PR #626, opened 2022-01-06
Construct prototype while training

karangrewal:
The `PrototypeContext` mixin has been augmented so that now it can do either of the following during training:

1. Create prototype contexts via task labels while training (just as before),
2. Construct prototype contexts via a clustering technique in an online way while training without need for tasks labels (new!).

There is also a corresponding config file for experiments that construct the prototype context while training. A few additional things to note:

- The attribute `contexts` (in the mixin) is a List containing a single Tensor if constructing the prototype context since lists are mutable, and we require `contexts` to be mutable since it is updated regularly by other functions,
- The clustering procedure requires float64, so the training process itself is slightly slower when constructing prototype contexts,
- There isn't any formal documentation of the exact clustering procedure in the mixin itself, and we can redirect to arXiv/bioRxiv papers.

Here are results from a stand-alone script that constructed prototypes (for continual learning on permutedMNIST) vs this Vernon implementation. The results are very similar, and averaged over 8 independent trials.

|          |  Raw  | Vernon |
|----------|-------|--------|
| 2 tasks  | 97.55 | 97.48  |
| 5 tasks  | 96.00 | 96.41  |
| 10 tasks | 94.32 | 94.53  |

### DIFF
--- a/packages/dendrites/src/nupic/research/frameworks/dendrites/mixins/prototype_context.py
+++ b/packages/dendrites/src/nupic/research/frameworks/dendrites/mixins/prototype_context.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2021, Numenta, Inc.  Unless you have an agreement
+# Copyright (C) 2022, Numenta, Inc.  Unless you have an agreement
 # with Numenta, Inc., for a separate license for this software code, the
 # following terms and conditions apply:
 #
@@ -21,7 +21,9 @@
 
 import abc
 
+import numpy as np
 import torch
+from scipy.stats import f
 
 from nupic.research.frameworks.dendrites import evaluate_dendrite_model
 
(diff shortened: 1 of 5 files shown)


packages/dendrites/src/nupic/research/frameworks/dendrites/mixins/prototype_context.py
```diff
@@ -21,7 +21,9 @@
 
 import abc
 
+import numpy as np
 import torch
+from scipy.stats import f
```
lscheinkman, line 26: Can you add this dependency to `packages/dendrites/setup.cfg` ?

lscheinkman (approved):

packages/dendrites/src/nupic/research/frameworks/dendrites/mixins/prototype_context.py
karangrewal: can I list `scipy.stats` as the dependency, or does it have to be `scipy`?

lscheinkman: just `spicy`

karangrewal: done.

subutai (approved):
👍

PR #632, opened 2022-02-08
Pure Python version of Spatial Pooler

abhi-iyer:
Pure Python 3 version of Spatial Pooler, no C++ dependencies. Cleans up a lot of the old spatial pooler code (removes unused methods, consolidates helper methods, fixes unclear logic, adds comments where ambiguous logic is used, etc.) and makes it more readable. 

Passes all tests. 
- Skipped test cases where `wrapAround=True` is used.
- Skipped `testUpdateBoostFactors` because `wrapAround=True` is the default setting for SP
- Skipped `testIsUpdateRound` and `updateDutyCycleHelper` because the contents of the test are already covered in other tests. Additionally, the helper methods that these tests cover are no longer standalone methods.

### DIFF
--- a/packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
+++ b/packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
@@ -1,0 +1,834 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2022, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy as np
+import itertools
+
+real_type = np.float32
+uint_type = np.uint32
+
+class SpatialPooler():
+    def __init__(self, 
+                 input_dims=(32, 32),
+                 minicolumn_dims=(64, 64),
+                 num_active_minicolumns_per_inh_area=10,
+                 local_density=-1.0,
+                 potential_radius=16,
+                 potential_percent=0.5,
+                 global_inhibition=False,
+                 stimulus_threshold=0,
+                 synapse_perm_inc=0.05,
+                 synapse_perm_dec=0.008,
+                 synapse_perm_connected=0.1,
+                 min_percent_overlap_duty_cycles=0.001,
+                 duty_cycle_period=1000,
+                 boost_strength=0.0,
+                 seed=-1,
+                 ):
+        '''
+        input_dims:                             dimensions of input vector. 
+                                                default ``(32, 32)``.
+
+        minicolumn_dims:                        dimensions of cortical minicolumns. 
+                                                default ``(64, 64)``.
+
+        num_active_minicolumns_per_inh_area:    sets number of minicolumns that remain on within a local inhibition area.
+                                                as minicolumns learn and grow their effective receptive fields, the inhibition radius will also grow. 
+                                                when this happens, net density of the number of active minicolumns *decreases* if this method is used.
+                                                default ``10.0``.
+
+        local_density:                          sets desired density of active minicolumns within local inhibition area. ensures that at most
+                                                N minicolumns remain on within a local inhibition area, 
+                                                where N = local_density * (total number of minicolumns in inhibition area). 
+                                                density of active minicolumns stays the same regardless of size of minicolumns' receptive fields. 
+                                                default ``-1.0``.
+
+        potential_radius:                       number of input bits visible to each minicolumn. large enough value means that every minicolumn can 
+                                                potentially connect to every input bit. this parameter defines a square/hypercube area: a minicolumn
+                                                will have a max square potential pool with sides of length (2 * potential_radius + 1).
+                                                default ``16``.
+
+        potential_percent:                      percent of inputs within a minicolumn's potential_radius that a minicolumn can be connected to. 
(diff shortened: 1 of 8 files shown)


subutai:
Thanks!  Quick question for @lscheinkman: should we create a whole new package just for the spatial pooler? Or should it be combined with the temporal memory code?

abhi-iyer:
We were thinking of keeping it as a new package, since the temporal memory code depends on building the `nupic.research.core` library. (Since this is Python only, it's not dependent on those C++ implementations.)

lscheinkman:
Ideally that would be the case, but as @abhi-iyer mentioned keeping as a separate package gives us more flexibility because it does not depend on anything other than `nupic.research`.  By putting in a separate package it also enable us to install it with our default conda env.

requirements.txt
```diff
@@ -11,9 +11,10 @@
...
 -e packages/self_supervised_learning
 -e packages/continual_learning
+-e packages/spatial_pooler
 
 # Install projects extra dependency
 # See `setup.cfg` for available extras.
 # -e .[imagenet]
 # -e .[gsc]
-# -e .[transformers]
+# -e .[transformers]
```
subutai, line 20: Is this correct?

subutai:
I see two issues with this:

1. It's really strange to create a package for a single file that has no additional requirements! (Note that we do want to eventually create a pytorch version of this, but that should also have no additional requirements over our regular pytorch install.)
2. At some point we will want to port over the version of the spatial pooler that uses the CPP bindings.  What happens then? Does that go into yet another package? Or into the columns package? Or would we incorporate it into this package so that all SP implementations are in the same package?

If we want to keep this file and package as no CPP, the naming, comments, and README should reflect that.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
```diff
@@ -0,0 +1,834 @@
...
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy as np
+import itertools
+
+real_type = np.float32
+uint_type = np.uint32
+
+class SpatialPooler():
```
subutai, line 28: Add comments that explain what the spatial pooler is to a naive reader. Simplest is to have one or two sentences explaining it and a second sentence pointing to the SP paper.

lscheinkman:
Thats a good point.
 I guess it would make sense to move SP to the columns package for now, so everything is consolidated in the same package and once we port the `cpp` code to `pytorch/python` we could just move it out of the packages into the `frameworks` directly and deprecate the package.

subutai:
To be clear, I think we'll end up with two implementations: one that uses pytorch and one that uses CPP bindings.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
```diff
@@ -0,0 +1,834 @@
...
+        self.minicolumn_dims = np.array(minicolumn_dims, ndmin=1)
+        self.num_inputs = np.prod(self.input_dims)
+        self.num_minicolumns = np.prod(self.minicolumn_dims)
+
+        assert (self.num_inputs > 0) and (self.num_minicolumns > 0)
+        assert (self.input_dims.size == self.minicolumn_dims.size)
+
+
+        ''' global or local inhibition '''
+        self.global_inhibition = global_inhibition
```
subutai, line 138: We should probably check that `global_inhibition` is either `True` or `False`.

abhi-iyer:
This might be poor logic on my part so please correct me if I'm wrong: if we keep it as a separate package with no dependencies other than `nupic.research`, we can update with PyTorch when necessary and still keep it in this standalone package. (i.e., if we do move to PyTorch, we wouldn't need the CPP bindings. PyTorch/NumPy implementations can both be in this `spatial_pooler` package.)

I forgot to make a README for this package -- I will do that.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
```diff
@@ -0,0 +1,834 @@
...
+            self.update_permanences_for_minicolumn(permanence, minicolumn_index, raise_perm=True)
+
+
+        self.inhibition_radius = 0
+        self.update_inhibition_radius()
+
+    
+    def compute(self, input_vector, learn, active_array):
+        '''
+        primary spatial pooler method. takes an input vector and outputs the indices of the active minicolumns. if learn is True, update the
```
subutai, line 236: Meta question: why do all sentences in comments start with lower case letters instead of the standard upper case?

subutai:
I think the only reason for packages is to deal with dependencies. If there are no dependencies, you can just create a spatial_pooler directory under Frameworks and put the file there. PyTorch and Numpy versions can both be in there. No separate install process will be needed - everyone gets access to the SP.

subutai (commented):
The code looks very clean. Some initial comments here.

requirements.txt
abhi-iyer: I removed the whitespace ending. GitHub I think shows it as I removed a line and then added it back.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
abhi-iyer: Wanted to keep the "Python" style of coding (probably just something I like to do) -- everything is lowercase, including variable names. No camel case.

abhi-iyer:
> I think the only reason for packages is to deal with dependencies. If there are no additional dependencies, you can just create a spatial_pooler directory under Frameworks and put the file there. PyTorch and Numpy versions can both be in there.
> 
> Tests would then go in nupic.research/tests/unit/frameworks
> 
> No separate install process will be needed - everyone gets access to the SP.

I still prefer keeping it as a separate package for 2 reasons:

1. If we use additional dependencies in the future, we can easily add it to the setup.cfg in the spatial_pooler package. An example of this is if we move this code to PyTorch and require new SparseMatrix bindings for the CPU/GPU. 

2. It's cleaner to have everything separated in a package (even though I understand that the original intent was to isolate dependencies). If the user chooses to use the spatial_pooler package, they can easily install it. 

What do you think? Maybe I'm off the mark.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
abhi-iyer: Done.

packages/spatial_pooler/src/nupic/research/frameworks/spatial_pooler/spatial_pooler.py
abhi-iyer: Done.

lscheinkman (approved):

PR #642, opened 2022-05-25
temporal memory wrapper

abhi-iyer:
ApicalTiebreakTemporalMemoryWrapper that doesn't have camelCase method names. 

The things we do to avoid flake8 errors 😅

### DIFF
--- a/packages/columns/src/nupic/research/frameworks/columns/temporal_memory_wrappers.py
+++ b/packages/columns/src/nupic/research/frameworks/columns/temporal_memory_wrappers.py
@@ -1,0 +1,68 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2022, Numenta, Inc.  Unless you have an agreement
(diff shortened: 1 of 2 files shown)


lscheinkman (approved):
Thank you 🥇

packages/columns/src/nupic/research/frameworks/columns/temporal_memory_wrappers.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+from nupic.research.frameworks.columns.apical_tiebreak_temporal_memory import (
+    ApicalTiebreakPairMemory,
+)
+
+
+class ApicalTiebreakPairMemoryWrapper(ApicalTiebreakPairMemory):
```
lscheinkman, line 27: Maybe it would be cleaner to just remove the camelCase from the original class

lscheinkman, line 27: Maybe it would be cleaner to just remove the camelCase from the original class

PR #653, opened 2024-03-06
Temporal Memory: Always sort predicted / predicted-active cells

mrcslws:
When setting the Columns paper code to use the `ApicalTiebreakTemporalMemory`, experiments that use feedback to L4 will occasionally send unsorted feedforward growth candidates into L2. I see this with `multi_column_synapse_sampling`. Many SparseMatrix methods expect these to be sorted, and they will fail quietly then cause undefined behavior, unless you have nupic.research.core compiled in Debug mode -- then an assert catches it.

Need to do some due diligence:
- Does this apply to the C++ `ApicalTiebreakTemporalMemory`?
- Add unit test covering this scenario. The scenario is: some minicolumns have cells predictive with only basal input, while other minicolumns have cells predictive with both basal and apical input. The list of predicted cells and predicted-active cells should remain sorted. (Make sure the test fails without this code change)


mrcslws:
Due diligence:
- This issue doesn't affect the temporal memory, it just affects consumers of the temporal memory who expect sorted outputs (which it *almost* always gives). I vaguely remember choosing this behavior on purpose, since the TM doesn't depend on sorted cells, but now it seems more pragmatic to take the slight perf hit and sort them.
- This doesn't affect the C++ ApicalTiebreakTemporalMemory, because the C++ version doesn't operate on batches of cells. There's never a point where the C++ version has a set of "cells in minicolumns where only basal support occurred" and another set of "cells in minicolumns where basal + apical support occurred". Instead, it loops over minicolumns in increasing order and handles them one by one.

PR #656, opened 2024-03-06
Port the Columns paper code to work in Python 3

mrcslws:
This change ports the [Columns paper](https://www.frontiersin.org/articles/10.3389/fncir.2017.00081/full) code and tests from htmresearch and gets it working in Python 3, using `nupic.research.core`. It resurrects some test dependencies from `nupic-legacy`.

To get this working without the Network API, I resurrected an old experimental change from 2017: https://github.com/numenta/htmresearch/pull/744 **"Experiment: Make L4L2Experiment use raw algorithms"**. Only some minor updates were needed after that change, along with some additions to the `nupic.research.core` bindings. (Most of the changed code in this PR comes from adapting code to pass flake8. Look at individual commits to see more useful diffs.)

I've verified the tests passed, and I've run each script. I haven't yet tried to reproduce the specific figures from the paper.


mrcslws:
Here are most of the figures from the paper, regenerated from this code. Keep in mind that the htmresearch code was updated after the Columns Paper with various side projects, so there may also be some subtle algorithm changes affecting results.

Most plots they are very similar to the original ones, if not identical. There's one exception: the convergence activity one (Figure 3) converges much faster in the single-column case. I suspect the paper used a different setup, e.g. with fewer locations or more objects. (Might be worth rerunning the htmpapers code with Python 2 in a Docker container?)

Compare against https://github.com/numenta/htmpapers/tree/master/frontiers/a_theory_of_how_columns_in_the_neocortex_enable_learning_the_structure_of_the_world

We have:

## Figure 3

image

image

image


## Figure 4

**4A**

image

**4B**

image

**4C**

image


## Figure 6

**6A**

image



**6B**

image

